PR #14, opened 2022-12-25
feat: add authority state

anonrig:


include/ada/url.h
```diff
@@ -94,23 +94,43 @@ namespace ada {
...
+      if (scheme == "file" && path.list_value.size() == 1 && checkers::is_normalized_windows_drive_letter(path.list_value[0])) {
+        return;
+      }
+
+      // Remove path’s last item, if any.
+      path.list_value.pop_back();
+    }
+
+    [[nodiscard]] std::optional<uint16_t> scheme_default_port() const {
+      return scheme::SPECIAL_SCHEME.find(scheme)->second;
```
addaleax, line 132: Do you use some specific form of assertion? It might be good to verify that `.find()` actually returns a result.

src/unicode.cpp
```diff
@@ -1,4 +1,5 @@
 #include <set>
+#include <locale>
```
addaleax, line 2: Are you using this?

src/parser.cpp
```diff
@@ -248,7 +248,7 @@ namespace ada::parser {
           // While c is an ASCII digit:
           while (unicode::is_ascii_digit(*pointer)) {
             // Let number be c interpreted as decimal number.
-            uint16_t number = static_cast<uint16_t>(*pointer);
+            auto number = static_cast<uint16_t>(std::strtoul(pointer, nullptr, 10));
```
addaleax, line 251: - I'm still not sure if the assumption that `std::string_view::iterator` is `const char*` is portable
- `input` is not guaranteed to be null-terminated
- `strtoul` has a number of behaviors that are probably incompatible with URL parsing (e.g. support for hex or octal digits, leading whitespace, etc.)
- You probably don't want to truncate the result here by casting it to `uint16_t` without regard for overflow

src/unicode.cpp
anonrig: Forgot to remove it. Thanks!

include/ada/url.h
anonrig: You're right. I missed this thinking `find()` returns `std::optional`, but value of SPECIAL_SCHEME is std::optional. I'm updating it now. Thanks!

src/parser.cpp
anonrig: Since pointer points to a ascii_digit, I think the best way to implement this is: `int number = *pointer - '0';`

anonrig: Since the pointer points to an ascii_digit, I think the best way to implement this is: `int number = *pointer - '0';`

PR #29, opened 2023-01-03
fix: move decoded domain to separate variable

anonrig:


anonrig:
This fixes the benchmarks. Merging it.

PR #45, opened 2023-01-10
perf: remove buffer concat from port state

anonrig:
Before: `8560ns`
After: `8144ns`

### DIFF
--- a/benchmarks/bench.cpp
+++ b/benchmarks/bench.cpp
@@ -28,8 +28,10 @@ std::string url_examples[] = {
     "https://www.reddit.com/?after=t3_zvz1ze",
     "https://www.reddit.com/login/?dest=https%3A%2F%2Fwww.reddit.com%2F",
     "postgresql://other:9818274x1!!@localhost:5432/otherdb?connect_timeout=10&application_name=myapp",
-    "http://192.168.1.1",
-    "http://[2606:4700:4700::1111]"};
+    "http://192.168.1.1", // ipv4
+    "http://[2606:4700:4700::1111]", // ipv6
+    "scheme:example.com:5432/path" // opaque-path
+};
 
 double url_examples_bytes =  []() {
     size_t bytes{0};
@@ -95,7 +97,7 @@ static void BasicBench_uriparser(benchmark::State& state) {
     }
   }
   uriFreeUriMembersA(&uri);
-  if(!is_valid) { std::cout << "invalid? " << std::endl; }
+  if(!is_valid) { std::cout << "uri-parser: invalid? " << std::endl; }
   state.counters["time/byte"] = benchmark::Counter(
 	        url_examples_bytes,
           benchmark::Counter::kIsIterationInvariantRate | benchmark::Counter::kInvert);
@@ -118,6 +120,7 @@ static void BasicBench_urlparser(benchmark::State& state) {
 }
 BENCHMARK(BasicBench_urlparser);
 
+// http_parser can't handle opaque path urls. this is a spec violation.
 static void BasicBench_http_parser(benchmark::State& state) {
   volatile bool is_valid{true};
   struct http_parser_url u;
@@ -127,7 +130,7 @@ static void BasicBench_http_parser(benchmark::State& state) {
       is_valid &= !http_parser_parse_url(url_string.data(), url_string.size(), 0, &u);
     }
   }
-  if(!is_valid) { std::cout << "invalid? " << std::endl; }
+  if(!is_valid) { std::cout << "http_parser: invalid? " << std::endl; }
   state.counters["time/byte"] = benchmark::Counter(
 	        url_examples_bytes,
           benchmark::Counter::kIsIterationInvariantRate | benchmark::Counter::kInvert);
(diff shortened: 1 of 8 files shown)


src/parser.cpp
```diff
@@ -1015,11 +1021,15 @@ namespace ada::parser {
...
+
             // If buffer is not the empty string, then:
-            if (!buffer.empty()) {
+            if (*port_start_pointer != pointer) {
+              std::string _buffer = std::string(*port_start_pointer, *port_end_pointer - *port_start_pointer);
+
               // Let port be the mathematical integer value that is represented by buffer in radix-10
               // using ASCII digits for digits with values 0 through 9.
-              long port = std::atol(buffer.c_str());
+              long port = std::atol(_buffer.c_str());
```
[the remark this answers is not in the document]
anonrig: What is the correct way of detecting uint32_t overflow?

src/parser.cpp
```diff
@@ -1015,14 +1018,17 @@ namespace ada::parser {
...
-            if (!buffer.empty()) {
+            if (std::distance(*port_start_pointer, pointer) > 0) {
               // Let port be the mathematical integer value that is represented by buffer in radix-10
               // using ASCII digits for digits with values 0 through 9.
-              long port = std::atol(buffer.c_str());
+              uint16_t port = helpers::get_port_fast(port_start_pointer.value(), port_end_pointer.value());
 
               // If port is greater than 216 − 1, validation error, return failure.
-              if (port > (1L<<16) - 1) {
+              // TODO: std::from_chars return 0 for buffer_overflow values
```
anonrig, line 1030: Adding a TODO until I find a better way for detecting uint16_t overflows.

src/parser.cpp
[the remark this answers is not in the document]
lemire: If you know that the string is all made of digits, then you can just check the string length. The maximal port value is 65535, so the string should not exceed 5 characters. Best thing might be to just checking the string prior to parsing...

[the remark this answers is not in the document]
lemire: But the C++ standard way would be this...

```C++

uint32_t get_port_fast(std::string_view::iterator begin, std::string_view::iterator end) noexcept {
    uint32_t port;
    std::string_view view(begin, end-begin);
    auto r = std::from_chars(view.data(), view.data() + view.size() , port);
    if(r.ec != std::errc()) { return -1; }
    return port;
}
```

`r.ec` can take different values depending on the nature of the error.

You can also return an optional from the above function if you'd like.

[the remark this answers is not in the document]
lemire: In all cases, if you go the  std::string_view, std::from_chars route, you get no uninlined function call and no allocation under GCC. You get something performant.  From what I can see, in general, GCC is unable to elide the temporary string, it seems to always create it... meaning that it allocates memory (possibly on stack) and then copy your content to it, then it has to later clean it up. It could also possibly throw.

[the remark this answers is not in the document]
anonrig: Updated the pull request. Used std::optional, since output is unsigned integer, `-1` is not available.

src/parser.cpp
anonrig: Resolved.

anonrig:
I think this is ready to get merged, but before doing it can you the run and share the benchmarks before after this pull request? I'm getting flaky results. @lemire

benchmarks/bench.cpp
```diff
@@ -28,8 +28,9 @@ std::string url_examples[] = {
     "https://www.reddit.com/?after=t3_zvz1ze",
     "https://www.reddit.com/login/?dest=https%3A%2F%2Fwww.reddit.com%2F",
     "postgresql://other:9818274x1!!@localhost:5432/otherdb?connect_timeout=10&application_name=myapp",
-    "http://192.168.1.1",
-    "http://[2606:4700:4700::1111]"};
+    "http://192.168.1.1", // ipv4
```
lemire, line 31: We are changing the benchmark. It makes tracking the performance over time difficult. It might be time to add *new* benchmarks. We can add new files like bench.cpp to benchmark specific cases.

anonrig: I was thinking about the same thing. Either we can separate them according to the host types (ipv4, ipv6, opaque etc.) or according to different contexts (For example: CURL benchmarks etc.)

anonrig:
I updated this branch, and the pull request description. These changes improves the performance around 400ns~ @lemire. It's ready for review & merge.

src/unicode.cpp
```diff
@@ -146,6 +146,11 @@ namespace ada::unicode {
     return input == "." || input == "%2e" || input == "%2E";
   }
 
+  // ipv4 character might contain 0-9 or a-f character ranges.
+  ada_really_inline constexpr bool is_ipv4_char(const char c) noexcept {
```
lemire, line 150: Maybe a better name: is_lowercase_hex

???

lemire (approved):

lemire:
This is good code. I would recommend merging even if it did not improve the performance. It is cleaner.

src/unicode.cpp
anonrig: Renaming

PR #93, opened 2023-01-22
perf: add percent_encode benchmarks

anonrig:
Some useful insights: UserInfo percent encode takes 3x slower

Screenshot 2023-01-22 at 12 06 21 PM


lemire (approved):
Let us merge this!!!

PR #107, opened 2023-01-23
feat: add is big endian detection

anonrig:
Fixes https://github.com/ada-url/ada/issues/97


anonrig:
Merging this. The changes are maintenance based.

PR #167, opened 2023-01-30
fix: remove leading/trailing control from href and port

anonrig:


anonrig:
This is only found because the original `json` dataset did not include all the test cases. For example: https://github.com/web-platform-tests/wpt/blob/master/url/url-setters-stripping.any.js

anonrig:
This is actually wrong. I'll open a new PR.

PR #173, opened 2023-01-31
Files do have null ports.

lemire:
Fixes: https://github.com/ada-url/ada/issues/42


anonrig:
@lemire Can you also add a test case for parsing a file path with a port?

anonrig (approved):

lemire:
@anonrig Test added.

PR #182, opened 2023-01-31
On common URLs, this should speed things up a bit in parse_host.

lemire:
Before... (bbc_bench)

```
----------------------------------------------------------------------------
Benchmark                  Time             CPU   Iterations UserCounters...
----------------------------------------------------------------------------
BasicBench_AdaURL       2872 ns         2872 ns       241453 speed=300.508M/s time/byte=3.3277ns time/url=319.089ns url/s=3.13392M/s
BasicBench_CURL         6501 ns         6501 ns       109582 speed=132.757M/s time/byte=7.53256ns time/url=722.289ns url/s=1.38449M/s
```

After... (bbc_bench)

```
----------------------------------------------------------------------------
Benchmark                  Time             CPU   Iterations UserCounters...
----------------------------------------------------------------------------
BasicBench_AdaURL       2508 ns         2508 ns       269107 speed=344.136M/s time/byte=2.90583ns time/url=278.637ns url/s=3.5889M/s
BasicBench_CURL         6486 ns         6486 ns       108238 speed=133.051M/s time/byte=7.51589ns time/url=720.691ns url/s=1.38756M/s
```

### DIFF
--- a/src/unicode.cpp
+++ b/src/unicode.cpp
@@ -24,6 +24,31 @@
 
 namespace ada::unicode {
 
+  constexpr bool to_lower_ascii(char * input, size_t length) noexcept {
+    auto broadcast = [](uint8_t v) -> uint64_t { return 0x101010101010101 * v; };
+    uint64_t broadcast_80 = broadcast(0x80);
+    uint64_t broadcast_Ap = broadcast(128 - 'A');
+    uint64_t broadcast_Zp = broadcast(128 - 'Z');
+    uint8_t non_ascii = 0;
+    size_t i = 0;
+
+    for (; i + 7 < length; i += 8) {
+      uint64_t word{};
+      memcpy(&word, input + i, sizeof(word));
+      non_ascii |= (word & broadcast_80);
+      word ^= (((word+broadcast_Ap)^(word+broadcast_Zp))&broadcast_80)>>2;
+      memcpy(input + i, &word, sizeof(word));
+    }
+    if (i < length) {
+      uint64_t word{};
+      memcpy(&word, input + i, length - i);
+      non_ascii |= (word & broadcast_80);
+      word ^= (((word+broadcast_Ap)^(word+broadcast_Zp))&broadcast_80)>>2;
+      memcpy(input + i, &word, length - i);
+    }
+    return non_ascii == 0;
+  }
+
   ada_really_inline constexpr bool has_tabs_or_newline(std::string_view user_input) noexcept {
     auto has_zero_byte = [](uint64_t v) {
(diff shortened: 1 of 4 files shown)


src/unicode.cpp
```diff
@@ -24,6 +24,31 @@
...
 
+  constexpr bool to_lower_ascii(char * input, size_t length) noexcept {
+    auto broadcast = [](uint8_t v) -> uint64_t { return 0x101010101010101 * v; };
+    uint64_t broadcast_80 = broadcast(0x80);
+    uint64_t broadcast_Ap = broadcast(128 - 'A');
+    uint64_t broadcast_Zp = broadcast(128 - 'Z');
+    uint8_t non_ascii = 0;
+    size_t i = 0;
+
+    for (; i + 7 < length; i += 8) {
```
anonrig, line 35: really nice. auto vectorization 👍

src/unicode.cpp
```diff
@@ -24,6 +24,31 @@
...
+      uint64_t word{};
+      memcpy(&word, input + i, sizeof(word));
+      non_ascii |= (word & broadcast_80);
+      word ^= (((word+broadcast_Ap)^(word+broadcast_Zp))&broadcast_80)>>2;
+      memcpy(input + i, &word, sizeof(word));
+    }
+    if (i < length) {
+      uint64_t word{};
+      memcpy(&word, input + i, length - i);
```
anonrig, line 44: Question: Why isn't this `sizeof(word)`?

src/url.cpp
```diff
@@ -449,14 +449,13 @@ namespace ada {
...
-      is_forbidden |= ada::unicode::is_forbidden_domain_code_point(c);
-      return (uint8_t((c|0x20) - 0x61) <= 25 ? (c|0x20) : c);}
-    );
+    std::string buffer = std::string(input);
+    // This next function checks that the result is ascii, but we are going to
+    // to check in anyhow with is_forbidden.
+    // bool is_ascii =
+    unicode::to_lower_ascii(buffer.data(), buffer.size());
+    uint8_t is_forbidden = 0;
+    for(char c : buffer) { is_forbidden |= ada::unicode::is_forbidden_domain_code_point(c); }
```
anonrig, line 458: would `std::any_of` a faster alternative?

src/unicode.cpp
lemire: This would be an overflow.

src/url.cpp
lemire: I think not. But I just rewrote it so that it is unrolled and thus typically faster.

anonrig (approved):

PR #185, opened 2023-01-31
fix: change the invalid function name for forbidden

anonrig:


anonrig:
Previously we were misusing the function name, even though the function implementation was correct.

lemire (approved):

PR #227, opened 2023-02-22
build: add code scanning workflow

anonrig:


anonrig:
Merging. Pretty straight forward, and non-blocking.

PR #264, opened 2023-03-06
feat: add more scheme related functions to url_base

anonrig:


lemire (approved):
This is quite decent. Making type what is effectively a publicly available variable has downsides: it allows someone to change it and doing so could leave the URL instance in a bad state. But that's fine with me.

PR #303, opened 2023-03-27
fix: use google clang-format style

anonrig:
Fixes https://github.com/ada-url/ada/issues/170


anonrig:
After this pull request, I think we should land `ada/idna` changes to main @lemire, what do you think?

Issue #205, opened 2023-02-04
Add workflow for releases


anonrig:
Creating a new github release can automatically trigger a workflow which can run `singleheadere/amalgate.py` and upload the zip file as an artifact to the github release. Would simplify the release process.

CarlosEduR:
merged, I think we can close this one

PR #312, opened 2023-03-29
Add workflow for releases

miguelteixeiraa:
ref: https://github.com/ada-url/ada/issues/205
Some of the workflows have:
```yml
on: [push, pull_request]
```
So they are also being triggered when creating a new tag. Should I move them to be something "like" this?:
```yml
on:
  push:
    branches:
      - "*"
    tags:
      - "!*"
  pull_request:
    branches:
      - main
``` 
I added a build&&test step before generating the assets, would that be enough or do we really need to run all those workflows?


.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+      - uses: actions/checkout@v3
+      - name: Amalgamation
+        run: cd singleheader && ./amalgamate.py
+
+      - name: Create Release && Upload Assets
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            singleheader/ada.cpp
+            singleheader/ada.h
```
anonrig, line 39: Can you also add `singleheader.zip`?

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: ctest --output-on-failure --test-dir build
+  release:
+    needs: build-and-test
+    runs-on: ubuntu-latest
+    if: ${{ needs.build-and-test.result == 'success' }}
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        
+    steps:
+      - uses: actions/checkout@v3
```
anonrig, line 30:
```suggestion
      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Amalgamation
+        run: cd singleheader && ./amalgamate.py
+
+      - name: Create Release && Upload Assets
```
anonrig, line 34: I think we should just upload assets to a newly created release, since we might also want to add some description on our own before releasing it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "*"
```
anonrig, line 6:
```suggestion
on:
  release:
    types: [published]
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+permissions:
+  contents: write
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    env:
+      CXX: clang++-14
+    steps:
+      - uses: actions/checkout@v3
```
anonrig, line 17:
```suggestion
      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
```

.github/workflows/release.yml
miguelteixeiraa: Do you think it'd be worth automating the release notes creation, or is this something that is better being a manual thing?

miguelteixeiraa:
Hello @anonrig @lemire
I still need to add some documentation, but I think it is ready for review :)
Here is an example of it working https://github.com/miguelteixeiraa/ada/actions/workflows/release.yml and https://github.com/miguelteixeiraa/ada/releases/tag/v4.5.0 (I'll remove those testing tags :sweat_smile:)

.github/workflows/release.yml
```diff
@@ -0,0 +1,83 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        type: string
```
anonrig, line 7: `required: true` is missing

.github/workflows/release.yml
```diff
@@ -0,0 +1,83 @@
...
+          path: ./tools/release/venv
+          key: venv-${{ hashFiles('tools/release/requirements.txt') }}
+      - name: Install dependencies
+        run: |
+          python3 -m venv venv
+          . ./venv/bin/activate
+          pip install -r requirements.txt
+      - name: Run tests
+        run: |
+          . ./venv/bin/activate
```
anonrig, line 38: Why do we have this `.` in the beginning?

.github/workflows/release.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Restore Python venv (Release script) from cache
+        uses: actions/cache@v3
+        with:
+          path: ./tools/release/venv
+          key: venv-${{ hashFiles('tools/release/requirements.txt') }}
+          restore-keys: venv-${{ hashFiles('tools/release/requirements.txt') }}
+        
+      - name: Update versions
+        run: |
+          . ./tools/release/venv/bin/activate
```
anonrig, line 58: Why do we have this `.` in here?

tools/release/lib/versions.py
```diff
@@ -0,0 +1,50 @@
...
+            print(line, end="")
+
+
+def update_doxygen_version(new_version, file_path):
+    with fileinput.FileInput(file_path, inplace=True) as doxygen:
+        for line in doxygen:
+            if line.strip().startswith("PROJECT_NUMBER         ="):
+                line = f'PROJECT_NUMBER         = "{new_version}"\n'
+
+            print(line, end="")
```
anonrig, line 50: Newline is missing from this file

anonrig (commented):
I added some comments. I think it's good work. After we merge this pull request and make sure it works, we should definitely move it to a different repo, probably under `ada/release` and expose it as a github action to be used for packages like simdutf and simdjson that @lemire maintains. It would help with mistakes like forgetting to update certain parts of the document.

.github/workflows/release.yml
```diff
@@ -0,0 +1,83 @@
...
+          git config --global user.name "GitHub Actions"
+
+          git add doxygen include/ada/ada_version.h CMakeLists.txt
+
+          git commit -m "New Release: ${{ env.NEXT_RELEASE_TAG }}"
+          git push
+
+      - name: "Create Release"
+        run: |
+          . ./tools/release/venv/bin/activate
```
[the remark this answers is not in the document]
miguelteixeiraa: > source and . are synonymous in Bash.

see https://askubuntu.com/a/25491
anyway "source" is more readable, I'll switch it

.github/workflows/release.yml
miguelteixeiraa: > source and . are synonymous in Bash.

see https://askubuntu.com/a/25491

miguelteixeiraa:
>  we should definitely move it to a different repo, probably under ada/release and expose it as a github action

Totally agree!

anonrig:
@miguelteixeiraa when can you address these issues? we need to release a new version, and depending on this pull request, we can try it

miguelteixeiraa:
> @miguelteixeiraa when can you address these issues? we need to release a new version, and depending on this pull request, we can try it

Todayyyyyy after lunch time GMT-3

miguelteixeiraa:
Done!

tools/release/lib/versions.py
```diff
@@ -0,0 +1,50 @@
...
+
+import fileinput
+import re
+
+
+def update_cmakelists_version(new_version, file_path):
+    inside_project = False
+    with fileinput.FileInput(file_path, inplace=True) as cmakelists:
+        for line in cmakelists:
+            if "set(ADA_LIB_VERSION" in line:
```
anonrig, line 11: For future: We need to make this configurable

tools/release/lib/versions.py
```diff
@@ -0,0 +1,50 @@
...
+                    inside_project = False
+            print(line, end="")
+
+
+def update_ada_version_h(new_version, file_path):
+    new_version_list = new_version.split(".")
+    with fileinput.FileInput(file_path, inplace=True) as ada_version_h:
+        inside_enum = False
+        for line in ada_version_h:
+            if "#define ADA_VERSION" in line:
```
anonrig, line 28: For future: We need to make this configurable

tools/release/update_versions.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python3
+
+import os
+import lib.versions as update_versions
+from lib.release import is_valid_tag
+
+
+WORK_DIR = os.path.dirname(os.path.abspath(__file__)).replace("/tools/release", "")
+
+ADA_VERSION_H = f"{WORK_DIR}/include/ada/ada_version.h"
```
anonrig, line 10: For future: We need to make this configurable

anonrig (commented):
@miguelteixeiraa Can you add a github workflow for running this script tests?

miguelteixeiraa:
> @miguelteixeiraa Can you add a github workflow for running this script tests?

Do you mean a separate workflow?
Currently it is set to run in the job [release-script-test](https://github.com/ada-url/ada/blob/b1c4791f676dc5688d1856c2a208204ea91cded5/.github/workflows/release.yml#L19), and the release only happen if the tests pass https://github.com/ada-url/ada/blob/b1c4791f676dc5688d1856c2a208204ea91cded5/.github/workflows/release.yml#L45

tools/release/create_release.py
```diff
@@ -0,0 +1,27 @@
...
+repo = g.get_repo(REPO_NAME)
+
+release_notes = release.contruct_release_notes(repo, NEXT_TAG)
+
+release.create_release(repo, NEXT_TAG, release_notes)
+
+release = repo.get_release(NEXT_TAG)
+release.upload_asset("singleheader/ada.cpp")
+release.upload_asset("singleheader/ada.h")
+release.upload_asset("singleheader/singleheader.zip")
```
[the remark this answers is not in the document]
miguelteixeiraa: Sure thing! I'll work on this!

anonrig:
> > @miguelteixeiraa Can you add a github workflow for running this script tests?
> 
> Do you mean a separate workflow? Currently it is set to run in the job [release-script-test](https://github.com/ada-url/ada/blob/b1c4791f676dc5688d1856c2a208204ea91cded5/.github/workflows/release.yml#L19), and the release only happen if the tests pass
> 
> https://github.com/ada-url/ada/blob/b1c4791f676dc5688d1856c2a208204ea91cded5/.github/workflows/release.yml#L45

You have added some tests with this pull request, to validate the output of the release script. Can we run them in a different github workflow to make sure we don't break anything?

miguelteixeiraa:
> > > @miguelteixeiraa Can you add a github workflow for running this script tests?
> > 
> > 
> > Do you mean a separate workflow? Currently it is set to run in the job [release-script-test](https://github.com/ada-url/ada/blob/b1c4791f676dc5688d1856c2a208204ea91cded5/.github/workflows/release.yml#L19), and the release only happen if the tests pass
> > https://github.com/ada-url/ada/blob/b1c4791f676dc5688d1856c2a208204ea91cded5/.github/workflows/release.yml#L45
> 
> You have added some tests with this pull request, to validate the output of the release script. Can we run them in a different github workflow to make sure we don't break anything?

Yeah, that makes sense. Done!

anonrig (approved):

PR #332, opened 2023-04-18
upgrade release script

anonrig:


miguelteixeiraa:
Interesting that it didn't need so much to make the python lint pass!

miguelteixeiraa (approved):

miguelteixeiraa:
Actually it looks like there is an error on line 26 https://github.com/ada-url/ada/blob/8be84228c3846092adaa8646492877f82fd70827/.github/workflows/lint_and_format_check.yml#L26

PR #340, opened 2023-04-18
Release process: Make actions send a PR instead of push into the main branch

miguelteixeiraa:
Now Github Actions should send a Pull Resquest with code changes in the following files:
```bash
doxygen
include/ada/ada_version.h
CMakeLists.txt
```
In order to this to work, we should provide to actions permission to create pull requests. It can be done through:
settings -> actions -> general -> Workflow permissions


.github/workflows/release.yml
```diff
@@ -68,15 +69,15 @@ jobs:
...
-          git push
+      - name: Create PR with code updates for new release
+        uses: peter-evans/create-pull-request@v5
+        with:
+          commit-message: "chore: release ${{ env.NEXT_RELEASE_TAG }}"
+          branch: "release/${{ env.NEXT_RELEASE_TAG }}"
+          title: "chore: release ${{ env.NEXT_RELEASE_TAG }}"
+          token: ${{ env.GITHUB_TOKEN }}
+          body: |
+            This pull PR updates the source code version to ${{ env.NEXT_RELEASE_TAG }}.
```
anonrig, line 80: `Create Release` script should run, only after the new pull request is merged.

anonrig (changes requested):

.github/workflows/release.yml
miguelteixeiraa: Hahaha fair enough

miguelteixeiraa:
The script should already consider co-authors now (https://github.com/ada-url/ada/issues/345).
I can try to fit the prepare and create workflows into a single one. What do you think @anonrig @lemire. 
Currently the workflow release_prepare.yaml should be triggered manually via workflow-dispatch with the target tag as input, and it will create a pull request with the code changes for the next version. When this pull request is merged, the workflow release_create.yml automatically creates the release and attach the artifacts.

.github/workflows/release-script-tests.yml
```diff
@@ -1,6 +1,7 @@
 name: Release Script Tests
 
 on:
+  workflow_call:
```
anonrig, line 4: Can you add a comment on top of this line with

```
# workflow_call is used to indicate that a workflow can be called by another workflow. 
```

.github/workflows/release-script-tests.yml
```diff
@@ -23,4 +24,4 @@ jobs:
       - name: Install dependencies
         run: pip3 install -r requirements.txt
       - name: Run tests
-        run: pytest -v
+        run: pytest -vv
```
anonrig, line 27: What is the difference between `-v` and `-vv`?

.github/workflows/release_create.yml
```diff
@@ -0,0 +1,47 @@
...
+          echo "NEXT_RELEASE_TAG=${NEXT_RELEASE_TAG}" >> $GITHUB_ENV
+
+      - name: Target release Tag
+        run: echo "New tag $NEXT_RELEASE_TAG"
+
+      - name: Amalgamation
+        run: ./singleheader/amalgamate.py
+      
+      - name: "Create release"
+        run: ./tools/release/create_release.py
```
anonrig, line 47: New line is missing at the end

tools/release/lib/release.py
```diff
@@ -1,12 +1,20 @@
...
+
 import re
+from typing import Optional, List, Set, Union, Type
+from github.PullRequest import PullRequest
+from github.GitRelease import GitRelease
+from github.Repository import Repository
 
 
-def is_valid_tag(tag):
+def is_valid_tag(tag: str) -> bool:
```
anonrig, line 10: Can you add a `ruff` rule to force using types in Python?

tools/release/lib/release.py
```diff
@@ -15,99 +23,180 @@ def create_release(repository, tag, notes):
...
+        ) from exp
+
 
+def get_sorted_merged_pulls(
+    pulls: List[PullRequest], last_release: Optional[GitRelease]
+) -> List[PullRequest]:
+    # Get merged pulls after last release
+    if not last_release:
+        return sorted(
+            (pull for pull in pulls if pull.merged and not pull.title.startswith("chore: release")),
```
anonrig, line 38: We need to also limit the author of this pull request title. If we choose to not use the script for any particular reason, this line would be wrong.

.github/workflows/release_prepare.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Update source code versions
+        run: ./tools/release/update_versions.py
+
+      - name: Ada Build
+        run: cmake -B build && cmake --build build
+      - name: Ada Test
+        run: ctest --output-on-failure --test-dir build
+
+      - name: Create PR with code updates for new release
+        uses: peter-evans/create-pull-request@v5
```
anonrig, line 44: Can you use an exact version for this? Like on line 30?

.github/workflows/release-script-tests.yml
miguelteixeiraa: Sure thing!

.github/workflows/release-script-tests.yml
miguelteixeiraa: This is for the level of verbosity https://docs.pytest.org/en/7.1.x/how-to/output.html?highlight=verbose#verbosity
There are higher levels btw -vvv, -vvvv.
But yeah, this does not need to be -vv, I'll switch to -v again.

tools/release/lib/release.py
miguelteixeiraa: Sure thing!

tools/release/lib/release.py
miguelteixeiraa: So let me see if i get it, It shouldn't be in here? :
https://github.com/ada-url/ada/blob/f393ed6e9fd32d289378949f61cc62e23ca268d5/.github/workflows/release_create.yml#L20

Or you mean really in release.py?

.github/workflows/release_prepare.yml
miguelteixeiraa: Sure thing!

tools/release/lib/release.py
anonrig: in `release.py` as well

miguelteixeiraa: Done! (for release.py and release_create.yml)

tools/release/lib/release.py
miguelteixeiraa: Just a small consideration, is this something we really want? I'm sorry I added this (type hints) in the code because it was growing and I just didn't want to be the only person understanding this. I wanted to make the code more readable and reduce the time required for new developers to make changes to it. I mean, I really like python with type hints and personally I would be very happy to have it as a rule, but I don't know, maybe someone doesn't feel that way. cc @lemire

miguelteixeiraa: Anyway, I added "ANN" (#flake8-annotations) in the mean time. Works like a charm!

miguelteixeiraa:
So currently it is working like this:
1. Manually trigger the workflow  **Release Prepare Workflow** passing the "next tag" as input.

     It will call the **release-script-test** workdlow to verify that the release script is still valid and will be able to make the necessary updates as expected.

     If the tests pass, it will run the update_versions.py script, which will update the versions of **ada_version.h**, **doxygen**, and **CMakeLists.txt** with the next release tag.

     After the updates to the code base, the workflow will run ada Build && Test and create a PR with the changes if everything is looks fine.

2. After step 1 is completed, a PR should appear in the repo with the name "chore: release vThe.New.Version".
Once this pull request is merged, it will trigger the **Release Create Workflow**, which will also run some tests, finally the Amalgamation, create the release notes and then the new github release.

anonrig (approved):

lemire (approved):

PR #358, opened 2023-04-26
Also update SOVERSION when release script runs

miguelteixeiraa:


tools/release/lib/versions.py
```diff
@@ -10,6 +10,8 @@ def update_cmakelists_version(new_version: str, file_path: str) -> None:
         for line in cmakelists:
             if "set(ADA_LIB_VERSION" in line:
                 line = re.sub(r"[0-9]+\.[0-9]+\.[0-9]+", new_version, line)
+            if "set(ADA_LIB_SOVERSION" in line:
+                line = re.sub(r"[0-9]+", new_version.split(".")[0], line)
```
anonrig, line 14: Can you add a test?

miguelteixeiraa: There is already a test for this. I just needed to update the expected result in here https://github.com/ada-url/ada/pull/358/files#diff-36665ed665c5a52f1f82c533f0774cbe7f686c77ca3121be185350f32cdd983cL10

The test is https://github.com/ada-url/ada/blob/main/tools/release/lib/tests/test_update_versions.py

lemire (approved):

PR #363, opened 2023-04-30
test: use gtest assertions

anonrig:


lemire (approved):
Looks very good to me.

PR #372, opened 2023-05-07
feat: add C bindings

anonrig:


tests/ada_c.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+TEST(ada_c, getters) {
+  void* url = ada_parse(
+      "https://username:password@www.google.com:8080/"
+      "pathname?query=true#hash-exists");
+
+  ASSERT_TRUE(ada_is_valid(url));
+
+  // TODO: Fix ada_get_origin returning invalid address.
+  //  ASSERT_EQ(convert_string(ada_get_origin(url)),
```
anonrig, line 27: @lemire do you have any suggestions for this?

anonrig:
By the way, should we expose the C API only through a build flag, or should we expose it always, like the current implementation does? @lemire

anonrig:
@miguelteixeiraa I've added `ada_c.h` to the uploaded assets since we are adding C API to the release. Can you review?

miguelteixeiraa:
Looks good!

PR #464, opened 2023-07-19
fix -Wunreachable-code-break warning

ckerr:
Fix a small `-Wunreachable-code-break` warning.

I found this a few steps downstream in Electron while working on Electron's Node.js roll, since Electron builds node with `-Werror,-Wunreachable-code-break` and Node.js uses ada :smile_cat:


anonrig:
Thanks for the PR @ckerr. Can you also open a PR to github.com/ada-url/idna for this fix?

anonrig (approved):

PR #470, opened 2023-08-01
feat: Implement url_search_params::has(name, value)

Zzzode:
According to https://url.spec.whatwg.org/#dom-urlsearchparams-has, we need to provide a `has` function that takes two arguments `name` and `value`.

### DIFF
--- a/fuzz/parse.cc
+++ b/fuzz/parse.cc
@@ -111,6 +111,10 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   }
   search_params.remove(source);
   search_params.remove(source, base_source);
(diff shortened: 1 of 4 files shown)


anonrig (approved):
Nice catch. Thank you for your contribution!

anonrig:
Can you also add this function to fuzz/parse.cc so we have some sort of fuzzing coverage?

Zzzode:
> Can you also add this function to fuzz/parse.cc so we have some sort of fuzzing coverage?

Sure, I'd be happy to handle this.

Zzzode:
@anonrig I don't have permission to merge, can you merge it?

anonrig:
Thank you!

PR #471, opened 2023-08-06
fix: remove unnecessary state assignment

anonrig:
The `state` is assigned inside the next state, and since `[[fallthrough]]` is used, we don't need to assign it.


lemire (approved):
It is likely that the compiler can figure it out… but sure!

PR #480, opened 2023-08-25
feat: add url host type attribute

anonrig:
This is one of the missing features comparing `ada-url` rust crate with `url` crate. Original implementation of `url` crate is available at: https://docs.rs/url/2.4.0/url/enum.Host.html


lemire (approved):
Very reasonable

PR #484, opened 2023-08-28
Avoid potentially unnecessary string reallocs.

ttsugriy:
In case `dest` is not created using a subset of the `input` `dest.reserve(input.size())` would result in increasing `dest`s size with likely reallocation. On top of being less efficient new version also generates significantly less asm with GCC 13. https://compiler-explorer.com/z/x3avEdWGM


anonrig (approved):

lemire (approved):
Let us merge.

PR #486, opened 2023-08-28
Avoid potentially unnecessary string reallocs.

ttsugriy:
In case `result` is not created using a subset of the `input` `result.reserve(input.size())` would result in increasing `result`'s size with likely reallocation.

It's similar to https://github.com/ada-url/ada/pull/484


anonrig (approved):

lemire (approved):
Sure.

PR #492, opened 2023-08-29
refactor: merge `ada_parse` and `ada_is_valid`

anonrig:
For environments where `ffi` has a cost, calling twice for every parsing operation reduces the performance by 15%. The following should in theory improve the performance.

PS: This is a breaking change. We didn't follow semver in C API until now, but should we?

### DIFF
--- a/include/ada_c.h
+++ b/include/ada_c.h
@@ -38,11 +38,16 @@ typedef struct {
 
 typedef void* ada_url;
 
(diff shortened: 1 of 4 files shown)


anonrig:
cc @lemire

anonrig:
Actually this resulted in worse performance for Rust. Closing.

PR #581, opened 2024-01-21
Fix check segment_result range(for node issue #51514)

ChenKS12138:
Fix an intermittent issue identified by @danfuzz in https://github.com/nodejs/node/issues/51514

# Reproudction

## Code Example

```c++
#include "ada.cpp"
#include "ada.h"
#include <iostream>

int main(int, char *[]) {
  auto url = ada::parse<ada::url>("https://1.1.1.256");
  if (!url) {
    std::cout << "failure" << std::endl;
    return EXIT_FAILURE;
  }
  url->set_protocol("http");
  std::cout << url->get_protocol() << std::endl;
  std::cout << url->get_host() << std::endl;
  return EXIT_SUCCESS;
}
```

## Expected
Raise an error and stop parse.

image

## Actual

Successfully parsed.

image

# Reason

This is an edge case, `segment_result` should not be equal to the size of remaining bits.

https://github.com/ada-url/ada/blob/4814d2f62f8c0c4545c39bcbaad3451eceda2ab3/src/url.cpp#L64-L83

### DIFF
--- a/tests/basic_tests.cpp
+++ b/tests/basic_tests.cpp
@@ -399,6 +399,6 @@ TYPED_TEST(basic_tests, nodejs_50235) {
 
 // https://github.com/nodejs/node/issues/51514
 TYPED_TEST(basic_tests, nodejs_51514) {
-  auto out = ada::parse<ada::url>("http://1.1.1.256");
+  auto out = ada::parse<TypeParam>("http://1.1.1.256");
   ASSERT_FALSE(out);
 }
--- a/tests/basic_tests.cpp
+++ b/tests/basic_tests.cpp
@@ -396,3 +396,9 @@ TYPED_TEST(basic_tests, nodejs_50235) {
   ASSERT_EQ(out->get_href(), "http://test.com:5/path?param=1");
   SUCCEED();
 }
+
+// https://github.com/nodejs/node/issues/51514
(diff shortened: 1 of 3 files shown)


anonrig (approved):

tests/basic_tests.cpp
```diff
@@ -396,3 +396,9 @@ TYPED_TEST(basic_tests, nodejs_50235) {
   ASSERT_EQ(out->get_href(), "http://test.com:5/path?param=1");
   SUCCEED();
 }
+
+// https://github.com/nodejs/node/issues/51514
+TYPED_TEST(basic_tests, nodejs_51514) {
+  auto out = ada::parse<ada::url>("http://1.1.1.256");
```
anonrig, line 402: We need to also fix and test this on url_aggregator

```suggestion
  auto out = ada::parse<TypeParam>("http://1.1.1.256");
```

anonrig (changes requested):

tests/basic_tests.cpp
ChenKS12138: Ok, I will fix it too.

ChenKS12138: I have fixed and tested it on `url_aggregator`.

image

lemire (approved):

anonrig (approved):

PR #609, opened 2024-03-10
build: use specific python version

anonrig:
Use specific python version (3.12) and install it with `setup-python` Github action.


bbayles (approved):

anonrig:
@lemire Can you take a look at this?

PR #608, opened 2024-03-10
refactor: update run-clangcldocker.sh

anonrig:
Small changes, nothing noteworthy

PR #610, opened 2024-03-17
Warning fixes and best practices applied

the-moisrex:
- Added .clang-tidy
- Added consts, lots of them
- removed do-whiles (https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-do-while.html)
- `static_cast` instead of c-style casts
- Signed and unsigned correctness of constant integers
- Removing `else` after `return`s
- Added some constexpr
- Added some curly braces
- Fixed some mismatched comments
- Used `std::any_of` instead of raw loops
- Removed some unnecessary `#incldue`s
- Renamed variables
- Used `auto` instead of a re-typing the obvious
- Used `.empty()` instead of checking the size
- Used some `std::move`s
- Moved some statements into `if-initialization`
- Removed redundant `std::string(...)`s
- Made `.to_string()` const
- Used `using` instead of `typedef`s
- Added some `[[nodiscard]]`s
- Removed some unnecessary `inline`s
- Other small changes...

I honestly got tired of typing `const` and `static_cast` :)


the-moisrex:
I can't reproduce the the [ASAN failed pipeline](https://github.com/ada-url/ada/actions/runs/8315863971/job/22754675593?pr=610), and my `clang-format`s are generating different results than the Linter check.

anonrig:
Can you run https://github.com/ada-url/ada/blob/main/tools/run-clangcldocker.sh to fix the formatter issues?

the-moisrex:
> Can you run https://github.com/ada-url/ada/blob/main/tools/run-clangcldocker.sh to fix the formatter issues?

It prints this, and nothing changes:

```
Trying to use docker
Using default tag: latest
latest: Pulling from kszonek/clang-format-15
Digest: sha256:eb991f6ac3cdaacfa9a5bbfec2fd7d4aa012ac8cffbc50bacb4f327b9d89f873
Status: Image is up to date for kszonek/clang-format-15:latest
docker.io/kszonek/clang-format-15:latest
Formatting [1/1] benchmarks/bbc_bench.cpp
benchmarks/bench.cpp
benchmarks/bench_search_params.cpp
benchmarks/benchmark_header.h
benchmarks/benchmark_template.cpp
benchmarks/competitors/servo-url/servo_url.h
benchmarks/model_bench.cpp
benchmarks/percent_encode.cpp
benchmarks/performancecounters/apple_arm_events.h
benchmarks/performancecounters/event_counter.h
benchmarks/performancecounters/linux-perf-events.h
benchmarks/wpt_bench.cpp
fuzz/parse.cc
include/ada.h
include/ada/ada_idna.h
include/ada/ada_version.h
include/ada/character_sets-inl.h
include/ada/character_sets.h
include/ada/checkers-inl.h
include/ada/checkers.h
include/ada/common_defs.h
include/ada/encoding_type.h
include/ada/expected.h
include/ada/helpers.h
include/ada/implementation.h
include/ada/log.h
include/ada/parser.h
include/ada/scheme-inl.h
include/ada/scheme.h
include/ada/serializers.h
include/ada/state.h
include/ada/unicode-inl.h
include/ada/unicode.h
include/ada/url-inl.h
include/ada/url.h
include/ada/url_aggregator-inl.h
include/ada/url_aggregator.h
include/ada/url_base-inl.h
include/ada/url_base.h
include/ada/url_components.h
include/ada/url_search_params-inl.h
include/ada/url_search_params.h
include/ada_c.h
singleheader/demo.c
singleheader/demo.cpp
src/ada.cpp
src/ada_c.cpp
src/ada_idna.cpp
src/checkers.cpp
src/helpers.cpp
src/implementation.cpp
src/parser.cpp
src/serializers.cpp
src/unicode.cpp
src/url-getters.cpp
src/url-setters.cpp
src/url.cpp
src/url_aggregator.cpp
src/url_components.cpp
tests/ada_c.cpp
tests/basic_fuzzer.cpp
tests/basic_tests.cpp
tests/from_file_tests.cpp
tests/url_components.cpp
tests/url_search_params.cpp
tests/wasm/wasm.cpp
tests/wpt_tests.cpp
tools/cli/adaparse.cpp
tools/cli/line_iterator.h
No such file or directory
```

lemire:
@anonrig I have verified that https://github.com/ada-url/ada/pull/608 broke the formatting routine for me.

the-moisrex:
> The sanitizer build issues might be unrelated. I am encountering other similar issues in other projects. I think it might be systemic.

I was suspected of this, but it would've been a compiler error if there was an issue with this!

image

lemire:
I was responding to this error:

Screenshot 2024-03-17 at 11 35 30 AM

It looks strange.

the-moisrex:
> @anonrig I have verified that #608 broke the formatting routine for me.

Reverted the script and reformatted again. I'm using clang-format version 17.0.6, and honestly, clang-format 15 differences is not that great, you might want to update clang-format in that script while fixing it.

The formatting pipeline works now.

the-moisrex:
> I was responding to this error:
> Screenshot 2024-03-17 at 11 35 30 AM
> 
> It looks strange.

Yes, I was banging my head around trying to reproduce it in a VM to see what's going on, but couldn't.

anonrig:
Can you split this pull request to multiple prs to reduce the change, to understand why ci is failing?

lemire:
Hmmm... The fuzzer errors might or might not be related. They have been flaky over different projects.

include/ada/helpers.h
```diff
@@ -209,23 +209,23 @@ inline int leading_zeroes(uint32_t input_num) noexcept {
...
+  constexpr static uint64_t table[] = {
       4294967296,  8589934582,  8589934582,  8589934582,  12884901788,
       12884901788, 12884901788, 17179868184, 17179868184, 17179868184,
       21474826480, 21474826480, 21474826480, 21474826480, 25769703776,
       25769703776, 25769703776, 30063771072, 30063771072, 30063771072,
       34349738368, 34349738368, 34349738368, 34349738368, 38554705664,
       38554705664, 38554705664, 41949672960, 41949672960, 41949672960,
       42949672960, 42949672960};
-  return int((x + table[int_log2(x)]) >> 32);
+  return static_cast<int>((x + table[int_log2(x)]) >> 32U);
```
lemire, line 228: Here you are 'correcting' 

```C++
return int((x + table[int_log2(x)]) >> 32);
```
and replacing it with 
```
static_cast<int>((x + table[int_log2(x)]) >> 32U);.
```

When would the more complicated version you propose be better/more correct. Can you give an example where the two expressions would give different results?

include/ada/scheme-inl.h
```diff
@@ -16,46 +16,50 @@ namespace ada::scheme {
...
 
 ada_really_inline constexpr bool is_special(std::string_view scheme) {
   if (scheme.empty()) {
     return false;
   }
-  int hash_value = (2 * scheme.size() + (unsigned)(scheme[0])) & 7;
+  auto const hash_value = static_cast<int>(
+      (2 * static_cast<int>(scheme.size()) + static_cast<unsigned>(scheme[0])) &
+      7U);
   const std::string_view target = details::is_special_list[hash_value];
```
lemire, line 33: You are replacing 
```C++
 int hash_value = (2 * scheme.size() + (unsigned)(scheme[0])) & 7;
```
with 
```
auto const hash_value = static_cast<int>(
       (2 * static_cast<int>(scheme.size()) + static_cast<unsigned>(scheme[0])) &
       7U);
```

Please explain when the two expressions would give different results.

include/ada/scheme-inl.h
```diff
@@ -16,46 +16,50 @@ namespace ada::scheme {
...
 constexpr uint16_t get_special_port(std::string_view scheme) noexcept {
   if (scheme.empty()) {
     return 0;
   }
-  int hash_value = (2 * scheme.size() + (unsigned)(scheme[0])) & 7;
-  const std::string_view target = details::is_special_list[hash_value];
+  auto const hash_value = static_cast<int>(
+      (2 * scheme.size() + static_cast<unsigned>(scheme[0])) & 7U);
+  std::string_view const target = details::is_special_list[hash_value];
   if ((target[0] == scheme[0]) && (target.substr(1) == scheme.substr(1))) {
```
lemire, line 43: You are replacing

```C++
const std::string_view target = details::is_special_list[hash_value];
```

with 

```
std::string_view const target = details::is_special_list[hash_value];
```

Can you explain why?

include/ada/scheme-inl.h
```diff
@@ -16,46 +16,50 @@ namespace ada::scheme {
...
 }
 constexpr uint16_t get_special_port(std::string_view scheme) noexcept {
   if (scheme.empty()) {
     return 0;
   }
-  int hash_value = (2 * scheme.size() + (unsigned)(scheme[0])) & 7;
-  const std::string_view target = details::is_special_list[hash_value];
+  auto const hash_value = static_cast<int>(
+      (2 * scheme.size() + static_cast<unsigned>(scheme[0])) & 7U);
+  std::string_view const target = details::is_special_list[hash_value];
```
lemire, line 42: You are replacing 

```
int hash_value = (2 * scheme.size() + (unsigned)(scheme[0])) & 7;
```

with

```
auto const hash_value = static_cast<int>(
       (2 * scheme.size() + static_cast<unsigned>(scheme[0])) & 7U);
```

When would the two expressions differ and why is the second one better/more correct?

include/ada/scheme-inl.h
```diff
@@ -16,46 +16,50 @@ namespace ada::scheme {
...
   }
+  return 0;
 }
-constexpr uint16_t get_special_port(ada::scheme::type type) noexcept {
-  return details::special_ports[int(type)];
+constexpr uint16_t get_special_port(ada::scheme::type const type) noexcept {
+  return details::special_ports[static_cast<int>(type)];
 }
-constexpr ada::scheme::type get_scheme_type(std::string_view scheme) noexcept {
+constexpr ada::scheme::type get_scheme_type(
```
lemire, line 51: You are replacing

```C++
constexpr uint16_t get_special_port(ada::scheme::type type) noexcept {
   return details::special_ports[int(type)];
 }
```

by 

```
constexpr uint16_t get_special_port(ada::scheme::type const type) noexcept {
   return details::special_ports[static_cast<int>(type)];
}
```

Why is the second expression better/more correct. When do they differ?

include/ada/checkers-inl.h
```diff
@@ -13,49 +13,57 @@
...
+  return (x >= '0') && (x <= '9');
+}
 
-constexpr char to_lower(char x) noexcept { return (x | 0x20); }
+constexpr char to_lower(char const x) noexcept {
+  return static_cast<char>(x | 0x20);
+}
 
-constexpr bool is_alpha(char x) noexcept {
+constexpr bool is_alpha(char const x) noexcept {
```
lemire, line 48: What is the purpose here of replacing

```C++
constexpr bool is_alpha(char x) noexcept {
```

with...
```
constexpr bool is_alpha(char const x) noexcept {
```

When do the expression differ, and why is the second expression better?

include/ada/checkers-inl.h
```diff
@@ -13,49 +13,57 @@
...
+  std::memcpy(&word0x, static_cast<void const*>(zero_x), 2);
+
   uint16_t two_first_bytes{};
   std::memcpy(&two_first_bytes, input.data(), 2);
   if (is_little_endian) {
-    two_first_bytes |= 0x2000;
+    two_first_bytes |= 0x2000U;
   } else {
-    two_first_bytes |= 0x020;
+    two_first_bytes |= 0x020U;
```
lemire, line 31: You are replacing `0x020` with `0x020U`. Why is this needed? Can you explain when the new code would give different results?

include/ada/url-inl.h
```diff
@@ -96,39 +100,39 @@ inline std::ostream &operator<<(std::ostream &out, const ada::url &u) {
...
 }
 
-inline void url::update_base_search(std::string_view input,
+inline void url::update_base_search(std::string_view const input,
                                     const uint8_t query_percent_encode_set[]) {
   query = ada::unicode::percent_encode(input, query_percent_encode_set);
 }
 
-inline void url::update_base_search(std::optional<std::string> input) {
+inline void url::update_base_search(std::optional<std::string> const &input) {
```
lemire, line 135: That's potentially an interesting change. By passing the std::optional by value, it is likely that we create a whole new copy of the string.

Good!!!!

include/ada/url-inl.h
```diff
@@ -96,39 +100,39 @@ inline std::ostream &operator<<(std::ostream &out, const ada::url &u) {
...
 
-inline void url::update_unencoded_base_hash(std::string_view input) {
+inline void url::update_unencoded_base_hash(std::string_view const input) {
   // We do the percent encoding
   hash = unicode::percent_encode(input,
                                  ada::character_sets::FRAGMENT_PERCENT_ENCODE);
 }
 
-inline void url::update_base_search(std::string_view input,
+inline void url::update_base_search(std::string_view const input,
```
lemire, line 130: Why would you ever pass a `const std::string_view` as opposed to a `std::string_view`?

include/ada/url-inl.h
```diff
@@ -223,7 +227,7 @@ ada_really_inline size_t url::parse_port(std::string_view view,
     return 0;
   }
   ada_log("parse_port: ", parsed_port);
-  const size_t consumed = size_t(r.ptr - view.data());
+  auto const consumed = static_cast<size_t>(r.ptr - view.data());
```
lemire, line 230: Why? You went from

```C++
const size_t consumed = size_t(r.ptr - view.data());
```

to...

```
auto const consumed = static_cast<size_t>(r.ptr - view.data());
```

I'd go in the other direction...

```C++
size_t consumed(r.ptr - view.data());
```

Less code, more readable.

include/ada/url.h
```diff
@@ -300,7 +297,7 @@ struct url : url_base {
   inline void update_base_search(std::string_view input);
   inline void update_base_search(std::string_view input,
                                  const uint8_t query_percent_encode_set[]);
-  inline void update_base_search(std::optional<std::string> input);
+  inline void update_base_search(std::optional<std::string> const &input);
```
lemire, line 300: I like this.

include/ada/url_aggregator-inl.h
```diff
@@ -51,17 +50,17 @@ inline void url_aggregator::update_base_authority(
     std::string_view password = input.substr(password_delimiter + 1);
 
     buffer.insert(components.protocol_end + diff, username);
-    diff += uint32_t(username.size());
+    diff += static_cast<uint32_t>(username.size());
```
lemire, line 53: Why would

```C++
diff += uint32_t(username.size());
```
be replaced by...

```
diff += static_cast<uint32_t>(username.size());
```

When is the latter more correct?

include/ada/url_aggregator-inl.h
```diff
@@ -104,10 +103,10 @@ inline void url_aggregator::update_unencoded_base_hash(std::string_view input) {
 }
 
 ada_really_inline uint32_t url_aggregator::replace_and_resize(
-    uint32_t start, uint32_t end, std::string_view input) {
-  uint32_t current_length = end - start;
-  uint32_t input_size = uint32_t(input.size());
-  uint32_t new_difference = input_size - current_length;
+    uint32_t const start, uint32_t const end, std::string_view const input) {
+  uint32_t const current_length = end - start;
+  auto const input_size = static_cast<uint32_t>(input.size());
```
lemire, line 108: You went from

```C++
uint32_t input_size = uint32_t(input.size());
```

to...

```
auto const input_size = static_cast<uint32_t>(input.size());
```

I'd go in the other direction and write...

```C++
uint32_t input_size(input.size());
```

Simpler, easier to read.

include/ada/url_aggregator-inl.h
```diff
@@ -155,7 +154,7 @@ inline void url_aggregator::update_base_hostname(const std::string_view input) {
 [[nodiscard]] ada_really_inline uint32_t
 url_aggregator::get_pathname_length() const noexcept {
   ada_log("url_aggregator::get_pathname_length");
-  uint32_t ending_index = uint32_t(buffer.size());
+  auto ending_index = static_cast<uint32_t>(buffer.size());
```
lemire, line 157: You went from

```C++
uint32_t ending_index = uint32_t(buffer.size());
```

to 

```
auto ending_index = static_cast<uint32_t>(buffer.size());
```

I'd go in the other direction and write...

```
uint32_t ending_index(buffer.size());

```

include/ada/url_aggregator-inl.h
```diff
@@ -404,7 +405,7 @@ inline void url_aggregator::append_base_username(const std::string_view input) {
     components.hash_start += difference;
   }
 #if ADA_DEVELOPMENT_CHECKS
-  std::string username_after = std::string(get_username());
+  std::string const username_after{get_username()};
```
lemire, line 408: The original is ugly and worth fixing...
```
std::string username_after = std::string(get_username());
```

But why do...

```
std::string const username_after{get_username()};
```

when you can do...

```C++
std::string username_after{get_username()};
```

which is nicer.

include/ada/url_aggregator-inl.h
```diff
@@ -493,7 +494,7 @@ inline void url_aggregator::append_base_password(const std::string_view input) {
   ADA_ASSERT_TRUE(!helpers::overlaps(input, buffer));
 #if ADA_DEVELOPMENT_CHECKS
   // computing the expected password.
-  std::string password_expected = std::string(get_password());
+  std::string password_expected{get_password()};
```
lemire, line 497: There you go. This is a nice change...
```
std::string password_expected = std::string(get_password());
```

is ugly.

This is much better...

```C++
std::string password_expected{get_password()};
```

include/ada/url_search_params-inl.h
```diff
@@ -194,20 +195,26 @@ inline bool url_search_params_iter<T, Type>::has_next() {
...
+  if (!has_next()) {
+    return std::nullopt;
+  }
   return params.params[pos++].first;
 }
 
 template <>
 inline std::optional<std::string_view> url_search_params_values_iter::next() {
-  if (!has_next()) return std::nullopt;
+  if (!has_next()) {
```
lemire, line 206: I agree. This is much better. I much prefer....

```C++
if(something) {
that
}
```

compared to the evil...

```
if(something) that
```

include/ada/url_search_params.h
```diff
@@ -21,7 +21,7 @@ enum class url_search_params_iter_type {
 template <typename T, url_search_params_iter_type Type>
 struct url_search_params_iter;
 
-typedef std::pair<std::string_view, std::string_view> key_value_view_pair;
+using key_value_view_pair = std::pair<std::string_view, std::string_view>;
```
lemire, line 24: using is a nice touch.

src/checkers.cpp
```diff
@@ -19,15 +19,15 @@ ada_really_inline ada_constexpr bool is_ipv4(std::string_view view) noexcept {
     }
     last_char = view.back();
   }
-  bool possible_ipv4 = (last_char >= '0' && last_char <= '9') ||
-                       (last_char >= 'a' && last_char <= 'f') ||
-                       last_char == 'x';
-  if (!possible_ipv4) {
+  if (bool const possible_ipv4 = (last_char >= '0' && last_char <= '9') ||
```
lemire, line 22: Why would you throw a const here?

src/checkers.cpp
```diff
@@ -88,37 +88,45 @@ ada_really_inline constexpr uint8_t path_signature(
...
   for (; i + 7 < input.size(); i += 8) {
-    accumulator |= uint8_t(path_signature_table[uint8_t(input[i])] |
-                           path_signature_table[uint8_t(input[i + 1])] |
-                           path_signature_table[uint8_t(input[i + 2])] |
-                           path_signature_table[uint8_t(input[i + 3])] |
-                           path_signature_table[uint8_t(input[i + 4])] |
-                           path_signature_table[uint8_t(input[i + 5])] |
-                           path_signature_table[uint8_t(input[i + 6])] |
-                           path_signature_table[uint8_t(input[i + 7])]);
+    accumulator |= static_cast<uint8_t>(
```
lemire, line 91: Here you have significantly harmed the readability.... This is not good:

```
 accumulator |= static_cast<uint8_t>(
         path_signature_table[static_cast<uint8_t>(input[i])] |
         path_signature_table[static_cast<uint8_t>(input[i + 1])] |
         path_signature_table[static_cast<uint8_t>(input[i + 2])] |
         path_signature_table[static_cast<uint8_t>(input[i + 3])] |
         path_signature_table[static_cast<uint8_t>(input[i + 4])] |
         path_signature_table[static_cast<uint8_t>(input[i + 5])] |
         path_signature_table[static_cast<uint8_t>(input[i + 6])] |
         path_signature_table[static_cast<uint8_t>(input[i + 7])]);
```

src/helpers.cpp
```diff
@@ -266,7 +267,7 @@ ada_really_inline size_t find_next_host_delimiter_special(
 }
 #elif ADA_SSE2
 ada_really_inline size_t find_next_host_delimiter_special(
-    std::string_view view, size_t location) noexcept {
+    std::string_view const view, size_t const location) noexcept {
```
lemire, line 270: Why would you even pass a const size_t or a const std::string_view... it burdens the function signature and brings no benefit.

src/helpers.cpp
```diff
@@ -623,12 +623,12 @@ ada_really_inline void parse_prepared_path(std::string_view input,
     // but dots must as appear as '.', and they cannot be encoded because
     // the symbol '%' is not present.
     size_t previous_location = 0;  // We start at 0.
-    do {
-      size_t new_location = input.find('/', previous_location);
+    for (;;) {
```
lemire, line 626: Why replace `do {} while (true)` with `for(;;;) {}`

How is that better?

lemire:
@the-moisrex 

There is a lot to like in there, but some that leaves me unconvinced.  This PR is too large for me to assess it all. For this reason alone, I don't think we should accept it (it is too difficult to review).

My proposal is that you break down this PR.

I suggest that you have the following PRs:

1. One PR where you fix the std::optional being passed by value. That's likely a mistake on our part, and a very good catch on your part. I would make it the first PR.
2. One where you make sure that if() is always followed by braces. That's quite good.
3. One where you simply our bad constructions (`std::string x = std::string(a)` and `uint32_t x = uint32_t(a)` are terrible, they need to be `std::string x(a)` and `uint32_t x(a)`.

There are other good changes, which should be distinct PRs. 

Ideally, your PRs should be separate.

Please do not add...

1. Use spurious unwarranted `static_cast`. If you use `static_cast`, then that must be to fix a bug. It is a noisy construction that I would rather avoid. Please note that it is fine to use it if it helps fix a bug. If it is only a matter of programming style, then I reject it.
2. Do not use unwarranted `const` qualifiers. If nothing else, we have too many const qualifiers. In some specific cases, it can be useful, but it almost never a good idea to add const where const was not present. I would rather that you remove const qualifiers than add them.
3. Do not add `auto` where it wasn't present unless it visibly improves readability.
4. If you want to replace `do{} while(...)` with `for(;;){}`, you have to convince us that it is better in our case. I know that you have the Core Guidelines but that's an argument by authority. We can about making the ada library better, not about abstract guidelines that may or may not help us.
5. In C/C++, by default an integer literal will be an int. That is fine. There is usually no need to add the `u` qualifier.

Regarding point 5, let us refer to the standard...

```
    7.3.1.8 Usual arithmetic conversions

        If both operands have the same type, then no further conversion is needed.
        Otherwise, if both operands have signed integer types or both have unsigned integer types, the operand with the type of lesser integer conversion rank is converted to the type of the operand with greater rank.
        Otherwise, if the operand that has unsigned integer type has rank greater or equal to the rank of the type of the other operand, then the operand with signed integer type is converted to the type of the operand with unsigned integer type.
        Otherwise, if the type of the operand with signed integer type can represent all of the values of the type of the operand with unsigned integer type, then the operand with unsigned integer type is converted to the type of the operand with signed integer type.
        Otherwise, both operands are converted to the unsigned integer type corresponding to the type of the operand with signed integer type.
```

I suggest that you only use unsigned literals when you have a reason to think, according to your interpretation of the standard, that the result might be wrong. It usually will not happen. 

Adding a suffix might be just noise that harms readability. It is best avoided.


> I honestly got tired of typing const and static_cast :)

Let me make clear that I deliberately avoid static_cast and I limit const qualifiers. This is very much by design. Let me restate the reason: it harms readability.

include/ada/url-inl.h
lemire: In the library, we put the const qualifier first... so it should be...

```
inline void url::update_base_search(const  std::optional<std::string> &input) {
```

include/ada/helpers.h
the-moisrex: `int(...)` is a c-style cast. It is impossible to find with a grep. The "intent" is to cast to int, why not express that intent?

The `U` suffix is there because **it's not a good idea to use signed integers in bitwise operations**, even though it has no effect here.

include/ada/scheme-inl.h
the-moisrex: I'm simply avoiding c-style casts, and intentionally express the casts in order to **avoid implicit casts**.

This should hail the same result as the previous code, if it doesn't it's a mistake from me.

include/ada/scheme-inl.h
the-moisrex: That's probably my clang-format re-arranging the qualifiers. I do prefer `const` to be after the type, but other than the style, it has no other value.

include/ada/scheme-inl.h
the-moisrex: It's the same thing as the one above, the use of `auto` is there because an additional cast is used to make sure we're avoiding implicit casts and everything explicitly said, and also we don't want to repeat ourselves either, thus the use of `auto`.

include/ada/checkers-inl.h
the-moisrex: Signed integers are not recommended to be used in bitwise operations.

include/ada/url-inl.h
the-moisrex: You could use clang-format for that and change everything in one go.

Here's the option that you need for clang-format in order to reorder the qualifiers:
image

include/ada/url-inl.h
the-moisrex: The same reason why we use `const` in variables. To help the compiler and also ourselves that the `input` is not gonna be used as a mutable variable during the body.

include/ada/url-inl.h
the-moisrex: Intent is more important than readability. We are "casting". implicit casting is a waste of time if we wanted to fix anything.

We SHOW the intent by using `static_cast`, and use `auto` to not repeat ourselves.

include/ada/url_aggregator-inl.h
the-moisrex: c-style cast looks better and more readable, I give you that, but it doesn't show the intent, and it certainly doesn't show which type of cast we need, and also it's trivially not searchable.

include/ada/url_aggregator-inl.h
the-moisrex: That would be an implicit cast hidden in it.

include/ada/url_aggregator-inl.h
the-moisrex: I'm sorry, but `const` has more value for me than simplicity. When I see a variable with no const, I expect it to change somewhere in the code.

src/checkers.cpp
the-moisrex: I feel you have something against `const` :)
If I don't see a const, I expect it to be modified later on.

src/checkers.cpp
the-moisrex: But the intent is more clear. Better way might be to have a separate function for this part:

```c++
path_signature_table[static_cast<uint8_t>(input[i + 1])]
```

But I don't agree with the use of c-style cast at all.

src/helpers.cpp
the-moisrex: Const there doesn't change the function signature, in fact, you shouldn't use const in the header files. But where the body of the function is, `const` helps to make the reader of the code (compiler or programmer) know that we're not gonna change this at all.

src/helpers.cpp
the-moisrex: You'd have to scroll 50/500/50000000000000000000000000 lines of code just to figure out that it's an infinite loop, then you have to scroll 50/500/50000000000000000000000000 lines of code, to go back where you started.

the-moisrex:
@lemire 

I did want to break this down in multiple PRs, but I thought it's going to be small changes, a few hours later, it became this.

- For the placement of `const`, I like the type to be the first thing I see, then other qualifiers, that's why I prefer `const` where I put them.
- I will not sacrifice readability (simply less characters in the code) over understandablity (if that's even a word). **If I see `const`, I know I'm not gonna be surprised, I won't have to look for the place the variable/parameter is changed or re-used**. `const` gives cleaner code reading ability. Of course, I do agree const should've been the default (like in Rust), but it's not, so more typing exercises for us C++ devs.
- `do{} while(true)` requires a lot of scrolling. So does `do{} while(...)`.
- `static_cast` is noisy, indeed. But it is explicit. Hidden implicit casts are hidden from the readers of the code as well until some extra focus and attention is paid.
- Adding `auto` in above cases are there because of the "no hidden implicit casts" strategy; when the whole thing is wrapped in a cast, there's no need to repeat the type of the variable.
- Adding suffixes to integers does in fact make it less readable, but it does make the reader understand the type and size of the integers used, and also signed integers should not be used in bitwise operations, it's confusing.

>> limit const qualifiers

I will not do such thing. it makes reading the code a horrible experience (I'm exaggerating obviously 🙃) because I have to keep looking for where the variable is modified and if I don't see it immediately, I feel I need to keep looking into functions that it's being passed into to see if any of them is modifying it or not; so it becomes a chasing practice and that's not fun when you want to work.

Separating the changes right now and make it multiple PRs seems like more work than reviewing them, let me know if it's a deal breaker.

lemire:
Thanks for your PR. I am closing this.

lemire:
For people reading this, I created several small PRs based on this large PR and they have been merged by @anonrig

PR #636, opened 2024-04-28
prevent missing python warning when format feature not requested

iamtimmy:
When consuming this library through fetch-content, cmake warns about missing python on my system.


anonrig (approved):

anonrig:
I think we should just remove it from cmake. For the time being, I'll merge this.

PR #638, opened 2024-04-28
chore: simplify clang-format and update ruff

anonrig:
Removes clang-format `cmake` configuration, updates clang-format and Ruff

Ref: https://github.com/ada-url/ada/pull/636


lemire (approved):
This PR is reformatting a dependency (expected).

PR #649, opened 2024-05-04
Add workflow file to run arm-based testing with sanitizers

CarlosEduR:
Fixes: #520

### DIFF
--- a/.github/workflows/aarch64.yml
+++ b/.github/workflows/aarch64.yml
@@ -1,0 +1,47 @@
+name: Ubuntu aarch64 (GCC 11)
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        shared: [ON, OFF]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: uraimo/run-on-arch-action@v2
+        name: Build
+        id: runcmd
+        with:
+          arch: aarch64
+          githubToken: ${{ github.token }}
+          distro: ubuntu_latest
+          install: |
+            apt-get update -q -y
+            apt-get install -y cmake make g++ ninja-build
+          run: |
+            cmake -DADA_SANITIZE=ON -DADA_DEVELOPMENT_CHECKS=ON -DBUILD_SHARED_LIBS=${{matrix.shared}} -G Ninja -B build
+            cmake --build build -j=2
+          env:
+            CXX: g++-12
+      - name: Test
+        run: ctest --output-on-failure --test-dir build


lemire:
Running tests.

CarlosEduR:
Seems it failed... I just pushed an update, let's see how it goes.

lemire:
Running...

CarlosEduR:
@lemire hmm, it is still failing... Off the top of your head, do you have any idea?
image

.github/workflows/aarch64.yml
```diff
@@ -0,0 +1,47 @@
...
+      - 'docs/**'
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+
+permissions:
+  contents: read
```
lemire, line 17: So you specified that files could not be written, correct?

lemire: I would just drop this permission step and see whether the permission issue resolved itself.

I am not quite sure why we specify permissions... :-/

.github/workflows/aarch64.yml
```diff
@@ -0,0 +1,47 @@
...
+          githubToken: ${{ github.token }}
+          distro: ubuntu_latest
+          install: |
+            apt-get update -q -y
+            apt-get install -y cmake make g++ ninja-build
+          run: |
+            cmake -DADA_SANITIZE=ON -DADA_DEVELOPMENT_CHECKS=ON -DBUILD_SHARED_LIBS=${{matrix.shared}} -G Ninja -B build
+            cmake --build build -j=2
+      - name: Test
+        run: ctest --output-on-failure --test-dir build
```
lemire, line 47: Alternatively, we could try removing `--output-on-failure` which might prevent ctest from creating logs, but I am not sure.

lemire:
It looks like a permission problem. I am not sure what the source of the issue is.

.github/workflows/aarch64.yml
CarlosEduR: Yes, actually I was more trying to follow the pattern I saw in other files. Also, the workflow file you shared with me as an example, it actually does not contain any permissions. I was wondering if adding permissions to write could my an issue.

lemire:
I have never seen this error:

```
CMake Error: Cannot open file for write: /home/runner/work/ada/ada/build/Testing/Temporary/LastTest.log.tmp
CMake Error: : System Error: Permission denied
```

lemire:
Let us run the tests...

.github/workflows/aarch64.yml
```diff
@@ -0,0 +1,44 @@
...
+          githubToken: ${{ github.token }}
+          distro: ubuntu_latest
+          install: |
+            apt-get update -q -y
+            apt-get install -y cmake make g++ ninja-build
+          run: |
+            cmake -DADA_SANITIZE=ON -DADA_DEVELOPMENT_CHECKS=ON -DBUILD_SHARED_LIBS=${{matrix.shared}} -G Ninja -B build
+            cmake --build build -j=2
+      - name: Test
+        run: ctest --test-dir build
```
anonrig, line 44: This needs to be inside the previous step just like the s360x workflow.

anonrig (changes requested):

.github/workflows/aarch64.yml
```diff
@@ -0,0 +1,43 @@
+name: Ubuntu aarch64 (GCC 11)
```
anonrig, line 1:
```suggestion
name: Ubuntu aarch64 (GCC 12)
```

anonrig (commented):
Can you also readd the permissions? It's important to limit the abilities of the workflow packages.

lemire:
@anonrig I suspect that the memory sanitizer is still busted. :-(

It finds problems inside the standard library.


(It is clearly unrelated to this PR.)

anonrig (approved):

lemire:
@CarlosEduR Yes. But at least we are now checking that things build, which is good.

PR #658, opened 2024-05-15
Add fuzzer for href setter and getter

CarlosEduR:


fuzz/href.cc
```diff
@@ -0,0 +1,33 @@
...
+
+#include "ada.cpp"
+#include "ada.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  FuzzedDataProvider fdp(data, size);
+  std::string source = fdp.ConsumeRandomLengthString(256);
+  std::string base_source = fdp.ConsumeRandomLengthString(256);
+
+  auto out_url = ada::parse<ada::url>(source);
```
anonrig, line 14: This won't work because 99.9999% of the time `out_url` will be false. Because the probability of the fuzzer to pass a correct URL is really-really low.

Ideally we should create a collection of tokens to be passed with the randomly generated input.

More information can be found from: https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/#dictionary

For example: `iajshduahfiuahsufdiasd` will not work at all, but if you surround it with `https://` and `.com` it will pass and still fuzz almost everything.

CarlosEduR:
Closing, it'll be pushed here: https://github.com/ada-url/ada/pull/657

PR #660, opened 2024-05-16
Add dictionary to run with fuzzers

CarlosEduR:
Relates to: https://github.com/ada-url/ada/pull/658


fuzz/url.dict
```diff
@@ -0,0 +1,2 @@
+"https://"
```
anonrig, line 1: I think you should add all types of protocols (available as special schemes in url spec)

anonrig (approved):

anonrig:
Is there anything that needs to be done to pass the dict to the fuzzer?

fuzz/url.dict
CarlosEduR: Ok, I'll do it

CarlosEduR:
> Is there anything that needs to be done to pass the dict to the fuzzer?

From what I was reading here, looks we just need this .options file.
image

I saw some other examples, just updating the build.sh to copy these files too and we should be fine. But let's see... 
https://github.com/google/oss-fuzz/blob/ad88a2e5295d91251d15f8a612758cd9e5ad92db/projects/expat/build.sh

fuzz/ada_fuzzer.options
```diff
@@ -0,0 +1,2 @@
+[libfuzzer]
+dict = url.dict
```
anonrig, line 2: Can you add ending missing newlines to each of these files? editorconfog extension for your ide should fix this

CarlosEduR: hmm, will update it. what ide do you use?

anonrig: zed and clion

CarlosEduR:
@anonrig would you like to set a max value here?
image

From examples, expat defined one but other projects were not.

anonrig:
We can do 1024

CarlosEduR:
@anonrig have you noticed any difference since last run? I haven't

anonrig:
> @anonrig have you noticed any difference since last run? I haven't

I haven't as well

PR #663, opened 2024-05-16
Use correct fuzzer options

anonrig:
Uses correct naming for the `options` file. The documentation states that it should have the target name. https://github.com/google/oss-fuzz/blob/master/docs/getting-started/new_project_guide.md#dictionaries

cc @CarlosEduR


lemire (approved):

CarlosEduR:
> Uses correct naming for the `options` file. The documentation states that it should have the target name. https://github.com/google/oss-fuzz/blob/master/docs/getting-started/new_project_guide.md#dictionaries
> 
> cc @CarlosEduR

Yes, I actually used a different one so it's shared by the different targets. @anonrig in this case, will it only be used by the fuzzer `parse`?

anonrig:
Yes, @CarlosEduR I assume that it didn't work like that. Let's see if the results change tomorrow.

PR #665, opened 2024-05-16
improve parse fuzzer

anonrig:
Fuzzer now runs on all setters, getters and `parse<type>`

cc @CarlosEduR


CarlosEduR:
Should we also use set_href here and I can just close my PR?

fuzz/parse.cc
```diff
@@ -9,28 +9,37 @@
...
+
+  if (parse_url_aggregator) {
+    length += parse_url_aggregator->get_href().size();
+  }
 
   /**
    * ada::parse<ada::url>
    */
-  auto out_url = ada::parse<ada::url>(source);
+  auto out_url = ada::parse<ada::url>("https://www.ada-url.com");
```
CarlosEduR, line 31: nice

PR #667, opened 2024-05-17
add memory sanitizer to cifuzz workflow

CarlosEduR:


CarlosEduR:
Found this issue: https://github.com/google/oss-fuzz/issues/11886
Seems it was already fixed, but the workflow file was also using an older version.

anonrig (approved):
nice

PR #669, opened 2024-05-17
fix: remove duplicate check on path_signature_table

anonrig:


lemire (approved):
Good catch. Note that a compiler will optimize this away. Still worth it to remove it to make the code more readable.

PR #670, opened 2024-05-17
perf: improve has_tabs_or_newline performance

anonrig:
This seems to slightly reduce the instruction count for `url_aggregator`, but please run local benchmarks to verify

Before:
- ada_url -> instructions/url=2.55753k
- ada_url_aggregator -> instructions/url=2.76077k

After:
- ada_url -> instructions/url=2.55742k
- ada_url_aggregator -> instructions/url=2.76044k

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -273,6 +273,7 @@ if(RUST_FOUND)
 
   # Check if servo-url target was created successfully
   if(TARGET servo-url)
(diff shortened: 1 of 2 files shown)


lemire (approved):
This is fine, but I bet you can do it with a lambda. Note that you are doing just the NEON routine, you could do the others as well.


I expect that this will not change performance. However, it can make the code more concise, which is good.

jasnell (approved):

CarlosEduR:
@anonrig good to go

PR #673, opened 2024-05-20
refactor(ada_c/ada_free_owned_string): remove unused code

pratikpc:
Setting parameters of an object passed by value [do nothing even at GCC Og level](https://godbolt.org/z/M8exqe8oh)


pratikpc:
The advantages are purely in readability.

There are no performance gains with optimizations turned on as the compiler can prove that the values are not going to be used whatsoever.

anonrig (commented):
These changes needs to be actually implemented in GitHub.com/ada-url/ada. The next time we update Ada, it will be reverted.

pratikpc:
Happens to us all LOL

jasnell (approved):

PR #674, opened 2024-05-23
perf: improve `ada::can_parse`

CarlosEduR:
Relates to: https://github.com/ada-url/ada/issues/377


CarlosEduR:
It's still like a POC, there might be a better place to put it, also even a better function name, I am still poking around.

CarlosEduR:
Benchmarks

main branch:
```
-------------------------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------------
BasicBench_AdaURL_CanParse   19704356 ns     19705721 ns           36 GHz=4.29878 cycle/byte=9.59934 cycles/url=833.791 instructions/byte=23.596 instructions/cycle=2.45809 instructions/ns=10.5668 instructions/url=2.04953k ns/url=193.96 speed=440.892M/s time/byte=2.26813ns time/url=197.008ns url/s=5.07594M/s
``` 

current branch:
```
-------------------------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------------
BasicBench_AdaURL_CanParse   14030751 ns     14029900 ns           49 GHz=4.274 cycle/byte=6.86874 cycles/url=596.613 instructions/byte=17.2687 instructions/cycle=2.5141 instructions/ns=10.7453 instructions/url=1.49995k ns/url=139.591 speed=619.255M/s time/byte=1.61484ns time/url=140.264ns url/s=7.12942M/s
```

CarlosEduR:
instead of creating a copy of it, we could also just add a parameter to the parse_url function, like store_values=true, and for the can_parse we set it to false.

CarlosEduR:
Screenshot from 2024-05-22 23-44-50

one example, `url.update_base_search` is called a few times, and it calls `ada::unicode::percent_encode`.

CarlosEduR:
I'd love to know what you guys think about it @anonrig @lemire

lemire:
I think we may want to keep...

```C++
template <typename result_type = ada::url_aggregator>
result_type parse_url(std::string_view user_input,
const result_type* base_url = nullptr);
```

as you have done, at least initially, so we don't mess up our signatures and it is less invasive. But we could add...

```C++
template <typename result_type, bool store_values>
result_type parse_url_impl(std::string_view user_input,
const result_type* base_url = nullptr);
```


And then the implementation of `parse_url` would just be...



```C++
template <typename result_type>
result_type parse_url(std::string_view user_input,
const result_type* base_url) { return parse_url_impl<result_type,true>(user_input); }
```

and `can_parse` would just call `parse_url_impl<result_type,false>`.

This should be nice because `store_values` will be a compile-time constant, so you can just branch on it, without any runtime cost.

Thoughts?

CarlosEduR:
nice!! it looks good to me

anonrig:
@lemire's suggestions looks really good, and this is amazing work @CarlosEduR. Thank you for doing this! I'm so excited!

CarlosEduR:
I appreciate the support, guys... I am working on it. 👨🏻‍💻

CarlosEduR:
with these small changes we go from this:

```
BasicBench_AdaURL_CanParse   19646497 ns     19645213 ns           36 GHz=4.29206 cycle/byte=9.6616 cycles/url=839.199 instructions/byte=23.3987 instructions/cycle=2.42183 instructions/ns=10.3946 instructions/url=2.03239k ns/url=195.523 speed=442.25M/s time/byte=2.26117ns time/url=196.403ns url/s=5.09157M/s
```

to this:
```
BasicBench_AdaURL_CanParse   13263878 ns     13263297 ns           53 GHz=4.27177 cycle/byte=6.5079 cycles/url=565.271 instructions/byte=16.6601 instructions/cycle=2.55998 instructions/ns=10.9357 instructions/url=1.44709k ns/url=132.327 speed=655.048M/s time/byte=1.52661ns time/url=132.6ns url/s=7.54149M/s
```

CarlosEduR:
I've noticed the `update_base_hostname` also cost a good amount for some cases, it's called by the `parse_host`. I was wondering if there we could avoid calling this update_base_hostname without affecting the can_parse/parse_host functionality. I'd like to hear what you guys think about it. 

For instance right here, by adding the `if(store_values)` around this code, we go from the last value `url/s=7.54149M/s` to `url/s=11.1449M/s`. Tests keep passing, but I am worried we are not covering other possible cases.

```
ada_log("HOST parsing ", host_view, " href=", url.get_href());
// Let host be the result of host parsing host_view with url is not
// special.
if (host_view.empty()) {
  url.update_base_hostname("");
} else if (!url.parse_host(host_view)) {
  return url;
}
ada_log("HOST parsing results in ", url.get_hostname(),
        " href=", url.get_href());

// Set url's host to host, and state to path start state.
state = ada::state::PATH_START;
```

CarlosEduR:
current perf report:

image

lemire:
> 've noticed the update_base_hostname also cost a good amount for some cases, it's called by the parse_host. I was wondering if there we could avoid calling this update_base_hostname without affecting the can_parse/parse_host functionality. I'd like to hear what you guys think about it.

In your example, it is called when `host_view.empty()` returns true. Then we call `url.update_base_hostname("")`. Is this really a hot function?

I am somewhat surprised that we are frequently setting the hostname to the empty string. But if it is common, then we can certainly try to optimize it.

CarlosEduR:
> In your example, it is called when `host_view.empty()` returns true. Then we call `url.update_base_hostname("")`. Is this really a hot function?
> 
> I am somewhat surprised that we are frequently setting the hostname to the empty string. But if it is common, then we can certainly try to optimize it.

I updated the code a little bit more and ran a benchmark to check, seems most of it is coming from the `parse_host`.
Code updated, it's only avoiding the `parse_host` function: 
```
if (host_view.empty()) {
  url.update_base_hostname("");
} else {
  if (store_values) {
    if (!url.parse_host(host_view)) {
      return url;
    }
  }
}
```
Benchmark result:
```
BasicBench_AdaURL_CanParse    9412687 ns      9412250 ns           75 GHz=4.25503 cycle/byte=4.56398 cycles/url=396.423 instructions/byte=11.8192 instructions/cycle=2.58966 instructions/ns=11.0191 instructions/url=1026.6 ns/url=93.1659 speed=923.062M/s time/byte=1083.35ps time/url=94.099ns url/s=10.6271M/s
```

src/parser.cpp
```diff
@@ -702,11 +706,13 @@ result_type parse_url(std::string_view user_input,
         } else {
           input_position = input_size + 1;
         }
-        if constexpr (result_type_is_ada_url) {
-          helpers::parse_prepared_path(view, url.type, url.path);
-        } else {
-          url.consume_prepared_path(view);
-          ADA_ASSERT_TRUE(url.validate());
+        if (store_values) {
```
anonrig, line 709:
```suggestion
        if constexpr (store_values) {
```

src/parser.cpp
```diff
@@ -539,19 +540,22 @@ result_type parse_url(std::string_view user_input,
...
-                             : ada::character_sets::QUERY_PERCENT_ENCODE;
-
-        // Percent-encode after encoding, with encoding, buffer, and
-        // queryPercentEncodeSet, and append the result to url's query.
-        url.update_base_search(helpers::substring(url_data, input_position),
-                               query_percent_encode_set);
-        ada_log("QUERY update_base_search completed ");
-        if (fragment.has_value()) {
-          url.update_unencoded_base_hash(*fragment);
+        if (store_values) {
```
anonrig, line 543:
```suggestion
        if constexpr (store_values) {
```

anonrig (commented):
Is there a reason to not use `constexpr`?

anonrig:
@CarlosEduR Can you also share before/after when you share your benchmark result? It's hard to understand the impact :-)

src/parser.cpp
CarlosEduR: No reason, I'll update it.

CarlosEduR:
> @CarlosEduR Can you also share before/after when you share your benchmark result? It's hard to understand the impact :-)

sure... currently:

before
```
BasicBench_AdaURL_CanParse   19646497 ns     19645213 ns           36 GHz=4.29206 cycle/byte=9.6616 cycles/url=839.199 instructions/byte=23.3987 instructions/cycle=2.42183 instructions/ns=10.3946 instructions/url=2.03239k ns/url=195.523 speed=442.25M/s time/byte=2.26117ns time/url=196.403ns url/s=5.09157M/s
```

after
```
BasicBench_AdaURL_CanParse   13263878 ns     13263297 ns           53 GHz=4.27177 cycle/byte=6.5079 cycles/url=565.271 instructions/byte=16.6601 instructions/cycle=2.55998 instructions/ns=10.9357 instructions/url=1.44709k ns/url=132.327 speed=655.048M/s time/byte=1.52661ns time/url=132.6ns url/s=7.54149M/s
```

src/implementation.cpp
```diff
@@ -47,16 +48,22 @@ std::string href_from_file(std::string_view input) {
...
 
 bool can_parse(std::string_view input, const std::string_view* base_input) {
-  ada::result<ada::url_aggregator> base;
+  ada::url_aggregator base_aggregator;
   ada::url_aggregator* base_pointer = nullptr;
+
   if (base_input != nullptr) {
-    base = ada::parse<url_aggregator>(*base_input);
-    if (!base) {
+    base_aggregator = ada::parser::parse_url_impl<ada::url_aggregator, false>(
```
anonrig, line 55: Let's not forget that, since we don't have `parse_url_impl<ada::url, false>` we won't get this function signature in any of the static releases.

Are we ok with only exposing this with `ada::url_aggregator`? @lemire

src/parser.cpp
```diff
@@ -1,18 +1,19 @@
...
-
-#include <limits>
 
 namespace ada::parser {
 
-template <class result_type>
-result_type parse_url(std::string_view user_input,
-                      const result_type* base_url) {
+template <class result_type, bool store_values>
+result_type parse_url_impl(std::string_view user_input,
```
anonrig, line 15: Can we keep the function name same? This is a breaking change.

CarlosEduR: I actually kept the old function `parse_url`, as Daniel suggested. 
image

Now the old `parse_url` calls the new one `parse_url_impl`, with the `store_values` as `true`.

src/parser.cpp
CarlosEduR: I just noticed and understood this comment, nice! I noticed you mentioned the `update_unencoded_base_hash`, and I'd like to confirm some other cases. For instance, `update_base_pathname` and `update_base_search` would be methods we want to avoid for the can_parse as well, right?

```
if constexpr (result_type_is_ada_url) {
  url.path = base_url->path;
  url.query = base_url->query;
} else {
  url.update_base_pathname(base_url->get_pathname());
  url.update_base_search(base_url->get_search());
}
url.update_unencoded_base_hash(*fragment);
return url;
```

src/implementation.cpp
lemire: I am not sure I understand the issue, but we should be very deliberate before breaking anything.

You do not want to get angry developers asking you why you pulled backward compatibility. 🥊

anonrig:
Is this ready to land? @CarlosEduR @lemire?

anonrig (approved):

CarlosEduR:
> Is this ready to land? @CarlosEduR @lemire?

Yeah, I can say it is. There might be more places to avoid allocation still, but I don't think they'll change the perf that much.

src/parser.cpp
```diff
@@ -539,19 +540,24 @@ result_type parse_url(std::string_view user_input,
...
+              url.is_special()
+                  ? ada::character_sets::SPECIAL_QUERY_PERCENT_ENCODE
+                  : ada::character_sets::QUERY_PERCENT_ENCODE;
+
+          // Percent-encode after encoding, with encoding, buffer, and
+          // queryPercentEncodeSet, and append the result to url's query.
+          url.update_base_search(helpers::substring(url_data, input_position),
+                                 query_percent_encode_set);
+          ada_log("QUERY update_base_search completed ");
+          if constexpr (store_values) {
```
anonrig, line 556: this if statement seems like a duplicate. there is already an if statement that checks for store values couple of lines ahove

CarlosEduR: True, updated!

CarlosEduR:
Last benchmark

before
```
BasicBench_AdaURL_CanParse   19773135 ns     19771058 ns           36 GHz=4.268 cycle/byte=9.65297 cycles/url=838.45 instructions/byte=23.3984 instructions/cycle=2.42396 instructions/ns=10.3455 instructions/url=2.03237k ns/url=196.45 speed=439.435M/s time/byte=2.27565ns time/url=197.661ns url/s=5.05916M/s
```

after
```
BasicBench_AdaURL_CanParse   13270228 ns     13269593 ns           53 GHz=4.29853 cycle/byte=6.50488 cycles/url=565.009 instructions/byte=16.6601 instructions/cycle=2.56117 instructions/ns=11.0093 instructions/url=1.44709k ns/url=131.442 speed=654.737M/s time/byte=1.52733ns time/url=132.663ns url/s=7.53791M/s
```

anonrig (approved):

PR #689, opened 2024-06-14
Add fuzzer for C wrapper

CarlosEduR:


fuzz/ada_c.c
```diff
@@ -0,0 +1,46 @@
...
+#include <string.h>
+
+int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  /**
+   * ada_c
+   */
+  ada_url out = ada_parse((char*)data, size);
+  bool is_valid = ada_is_valid(out);
+
+  if (out) {
```
anonrig, line 15:
```suggestion
  if (is_valid) {
```

fuzz/ada_c.c
```diff
@@ -0,0 +1,46 @@
...
+    ada_get_hash(out);
+    ada_get_host(out);
+    ada_get_host_type(out);
+    ada_get_hostname(out);
+    ada_get_href(out);
+    ada_get_origin(out);
```
anonrig, line 21: you should free this, or it will leak and give false result

fuzz/ada_c.c
```diff
@@ -0,0 +1,46 @@
...
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  /**
+   * ada_c
+   */
+  ada_url out = ada_parse((char*)data, size);
```
anonrig, line 12: this needs to be freed as well

CarlosEduR: Code added freeing them

CarlosEduR:
I plan to call the setters as well, then will make it ready for review.

PR #727, opened 2024-09-02
Fix setters to return correct boolean value

CarlosEduR:
Fix: #719


CarlosEduR:
I noticed we have some other cases here just returning true as well, would it be correct to update all of them to return false instead?

`url::parse_scheme`
```C++
auto parsed_type = ada::scheme::get_scheme_type(input);
bool is_input_special = (parsed_type != ada::scheme::NOT_SPECIAL);
/**
 * In the common case, we will immediately recognize a special scheme (e.g.,
 *http, https), in which case, we can go really fast.
 **/
if (is_input_special) {  // fast path!!!
  if (has_state_override) {
    // If url's scheme is not a special scheme and buffer is a special scheme,
    // then return.
    if (is_special() != is_input_special) {
      return true;
    }
    // If url includes credentials or has a non-null port, and buffer is
    // "file", then return.
    if ((has_credentials() || port.has_value()) &&
        parsed_type == ada::scheme::type::FILE) {
      return true;
    }
    // If url's scheme is "file" and its host is an empty host, then return.
    // An empty host is the empty string.
    if (type == ada::scheme::type::FILE && host.has_value() &&
        host.value().empty()) {
      return true;
    }
  }

  type = parsed_type;
```

lemire:
@CarlosEduR Important work!!!

Yes, I think that in the code sample you have found, it should return false in all the error cases.

anonrig:
@CarlosEduR is this ready to review? i see it as draft?

CarlosEduR:
@anonrig now it is...

lemire (approved):
Looks like it improves ada.

anonrig:
@CarlosEduR tests seem to be failing with the most recent changes

lemire:
@anonrig 

> tests seem to be failing with the most recent changes

Which tests?

anonrig:
Github UI seems to be failing for me

image

anonrig (approved):

lemire:
I am merging.

CarlosEduR:
desktop web looks ok
image

PR #730, opened 2024-09-03
simplify release generator

anonrig:
Instead of generating our release notes, we are now using `generate_release_notes=True` option in `pygithub`


tools/release/lib/release.py
```diff
@@ -1,201 +1,21 @@
...
+from github import Repository, GitRelease
 
 
 def is_valid_tag(tag: str) -> bool:
     tag_regex = r'^v\d+\.\d+\.\d+$'
     return bool(re.match(tag_regex, tag))
 
 
-def create_release(repository: Repository, tag: str, notes: str) -> Union[None, Type[Exception]]:
+def create_release(repository: Repository, tag: str) -> GitRelease:
```
CarlosEduR, line 12: I believe this would be: `GitRelease.GitRelease`

anonrig: It seems `from github import Repository, GitRelease` also works :-)

PR #733, opened 2024-09-06
update ada::idna to c++20

anonrig:
Updates idna code to `main` branch of the project


lemire (approved):
Notice how our CI turns green.

So satisfying.

PR #754, opened 2024-10-07
remove unnecessary std::string_view creation

anonrig:
Removes unnecessary string_view creation and adds likely and unlikely to several branches.


lemire (approved):
It is all fine.

I would be careful generally with likely and unlikely markers. The compiler might not do what you expect.

In this case, it is unlikely to matter either way.

PR #788, opened 2024-11-30
use more c++20 functions

anonrig:
Let's use more C++20 functionality.


lemire (approved):
Hopefully, this is performance neutral. I expect so. :-)

The new code is simpler.

PR #804, opened 2024-12-20
Improved Memory Management Proper Resource Cleanup and Freeing Allocated Memory

Shivam7-1:
This PR improves memory management by ensuring all dynamically allocated resources are properly freed. It adds early exits with cleanup for invalid parsing and URL checks, ensures out and out_get_origin are freed in all exit paths, and prevents memory leaks, enhancing the robustness and efficiency of the fuzzer code.

### DIFF
--- a/fuzz/ada_c.c
+++ b/fuzz/ada_c.c
@@ -10,52 +10,62 @@ int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
    * ada_c
    */
   ada_url out = ada_parse((char*)data, size);
-  bool is_valid = ada_is_valid(out);
-
-  if (is_valid) {
-    ada_set_href(out, (char*)data, size);
-    ada_set_host(out, (char*)data, size);
-    ada_set_hostname(out, (char*)data, size);
-    ada_set_protocol(out, (char*)data, size);
-    ada_set_username(out, (char*)data, size);
-    ada_set_password(out, (char*)data, size);
-    ada_set_port(out, (char*)data, size);
-    ada_set_pathname(out, (char*)data, size);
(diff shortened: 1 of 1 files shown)


Shivam7-1:
Hii @anonrig Could you please Review This PR whenever you are free :)
Thanks

fuzz/ada_c.c
```diff
@@ -10,52 +10,62 @@ int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
...
+  if (!out) {
+    return 0;
+  }
+
   bool is_valid = ada_is_valid(out);
+  
+  if (!is_valid) {
+    ada_free(out);
+    return 0;
+  }
```
anonrig, line 23: This is not ideal because if the result has failed to parse, we still want to continue to the remaining of the fuzzer. Particularly, if ada_parse fails, we don't test ada_can_parse

anonrig (changes requested):
This would reduce our fuzzer coverage.

Shivam7-1:
Hii 
I had done some changes above  that the fuzzer continues even if `ada_parse` fails or the URL is invalid, improving test coverage. Instead of returning immediately, memory is properly managed by freeing allocated resources, but the fuzzer still proceeds to test `ada_can_parse` and other functions. 
This avoids prematurely ending tests while ensuring correct memory cleanup for invalid inputs.

fuzz/ada_c.c
Shivam7-1: Thank you for the feedback! I have updated the code to ensure that the fuzzer continues even if ada_parse fails or the URL is invalid. Instead of returning early the code now proceeds to test ada_can_parse

PR #813, opened 2024-12-27
Remove barrel files

anonrig:
Work in progress.


anonrig:
I'll open a new one once URLPattern lands.

PR #814, opened 2024-12-28
fix invalid regex pattern causing test failures

mertcanaltin:
replaced invalid regex pattern " * " with " .* "
ensured compatibility with ecmascript syntax


mertcanaltin:
my contribution here was very small so I added this as a comment, I'm closing this place

PR #820, opened 2025-01-02
feat: implement parsing for expected_match field

mertcanaltin:
I added parsing logic for expected_match to handle strings, null values, and objects safely

I’ve opened a PR on top of Yagiz’s (@anonrig) PR. Only the last commit is mine. I’d greatly appreciate it if you could take a look @lemire


mertcanaltin:
I accidentally opened maine and I'm turning it off ❤️

PR #821, opened 2025-01-02
[WIP] feat: implement parsing for expected_match field

mertcanaltin:
I added parsing logic for expected_match to handle strings, null values, and objects safely

@lemire @anonrig


tests/wpt_urlpattern_tests.cpp
```diff
@@ -375,7 +375,30 @@ TEST(wpt_urlpattern_tests, urlpattern_test_data) {
...
         // - {} // response here.
         auto input_value = parse_inputs_array(inputs);
-        // TODO: Parse "expected_match" field here.
+
+        ondemand::value expected_match_value;
+        if (main_object["expected_match"].get(expected_match_value)) {
+          if (expected_match_value.type() ==
+              simdjson::ondemand::json_type::string) {
+            std::string_view expected_match_str;
+            if (expected_match_value.get_string().get(expected_match_str)) {
```
anonrig:
```suggestion
            ASSERT_FALSE(expected_match_value.get_string().get(expected_match_str));
```

anonrig: We can replace this if statement with this. no need to check once again.

tests/wpt_urlpattern_tests.cpp
```diff
@@ -375,7 +375,30 @@ TEST(wpt_urlpattern_tests, urlpattern_test_data) {
...
+            }
+          } else if (expected_match_value.type() ==
+                     simdjson::ondemand::json_type::null) {
+            std::cout << "Expected match is null." << std::endl;
+          } else if (expected_match_value.type() ==
+                     simdjson::ondemand::json_type::object) {
+            auto expected_match_obj = expected_match_value.get_object();
+            std::cout << "Expected match is an object." << std::endl;
+          } else {
+            std::cout << "Unexpected type for expected_match." << std::endl;
```
anonrig, line 397:
```suggestion
            FAIL() << "Unexpected type for expected_match.";
```

tests/wpt_urlpattern_tests.cpp
```diff
@@ -375,7 +375,30 @@ TEST(wpt_urlpattern_tests, urlpattern_test_data) {
...
+          }
+        } else {
+          std::cout << "Expected match field not found or invalid."
+                    << std::endl;
```
anonrig, line 401:
```suggestion
```

tests/wpt_urlpattern_tests.cpp
mertcanaltin: thanks, I changed

mertcanaltin:
Hi everyone,

While working on an improvement, I encountered this error during debugging:
```JSON error: OUT_OF_ORDER_ITERATION: Objects and arrays can only be iterated when they are first encountered.```

The issue occurs with this line:

```c
auto key = field.key().value();
```
Could this be because the url_pattern_result method tries to reuse the JSON object?

Issue #826, opened 2025-01-05
set_port returns true for invalid value


anonrig:
Ref: https://github.com/ada-url/ada/pull/785/commits/e958292de455ea497ad58495f9a092cebf6c88d5

Calling set_port "invalid80" on "fake://dummy.test" returns true, but it should return false

anonrig:
cc @mertcanaltin, this is a good first issue that you can open to main branch

mertcanaltin:
thank you I will take care of it right away 🚀

PR #829, opened 2025-01-05
fix: Improve `set_port` validation and `get_host` behavior

mertcanaltin:
fix Ensure `set_port` correctly handles invalid port values

#826

### DIFF
--- a/src/ada_c.cpp
+++ b/src/ada_c.cpp
@@ -290,7 +290,19 @@ bool ada_set_port(ada_url result, const char* input, size_t length) noexcept {
   if (!r) {
     return false;
   }
-  return r->set_port(std::string_view(input, length));
+
+  std::string_view port_input(input, length);
+
+  if (port_input.find_first_not_of("0123456789") != std::string_view::npos) {
+    return false;
+  }
+
+  int port = std::stoi(std::string(port_input));
+  if (port < 1 || port > 65535) {
+    return false;
+  }
+
+  return r->set_port(port_input);
 }
 
 bool ada_set_pathname(ada_url result, const char* input,
--- a/src/ada_c.cpp
+++ b/src/ada_c.cpp
@@ -290,19 +290,7 @@ bool ada_set_port(ada_url result, const char* input, size_t length) noexcept {
   if (!r) {
     return false;
   }
-
-  std::string_view port_input(input, length);
-
-  if (port_input.find_first_not_of("0123456789") != std::string_view::npos) {
-    return false;
-  }
-
-  int port = std::stoi(std::string(port_input));
-  if (port < 1 || port > 65535) {
-    return false;
-  }
-
-  return r->set_port(port_input);
+  return r->set_port(std::string_view(input, length));
 }
 
 bool ada_set_pathname(ada_url result, const char* input,
--- a/src/url_aggregator.cpp
+++ b/src/url_aggregator.cpp
@@ -298,6 +298,13 @@ bool url_aggregator::set_port(const std::string_view input) {
   // Revert changes if parse_port fails.
   uint32_t previous_port = components.port;
   parse_port(trimmed);
+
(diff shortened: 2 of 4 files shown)


src/url_aggregator.cpp
```diff
@@ -298,6 +298,13 @@ bool url_aggregator::set_port(const std::string_view input) {
   // Revert changes if parse_port fails.
   uint32_t previous_port = components.port;
   parse_port(trimmed);
```
anonrig, line 300: I think this check needs to be done inside parse port. Right now, you're doing the same work twice.

tests/basic_tests.cpp
```diff
@@ -17,6 +17,7 @@ TYPED_TEST(basic_tests, insane_url) {
   ASSERT_EQ(r->get_hostname(), "");
   ASSERT_EQ(r->get_port(), "");
   ASSERT_EQ(r->get_pathname(), "@EEEEEEEEEE");
+  ASSERT_FALSE(url->set_port("invalid80"));
```
anonrig, line 20: This is perfect but your test will fail on Ada::url. You changed url_aggregator set_port implementation but not the other one.

anonrig, line 20: I recommend running tests before opening a PR

```suggestion
  ASSERT_FALSE(r->set_port("invalid80"));
```

mertcanaltin: thank you I ran the tests but I got passed I wonder if I did the build wrong

I usedfirst this command: 
```
cmake -B build -DADA_TESTING=ON -DCMAKE_BUILD_TYPE=Debug                  
cmake --build build
```

after I used then this command for all tests:
`ctest --output-on-failure --test-dir build`

anonrig: > ctest --output-on-failure --test-dir build

What's the output of this command for you?

src/ada_c.cpp
```diff
@@ -290,7 +290,19 @@ bool ada_set_port(ada_url result, const char* input, size_t length) noexcept {
...
+  if (port_input.find_first_not_of("0123456789") != std::string_view::npos) {
+    return false;
+  }
+
+  int port = std::stoi(std::string(port_input));
+  if (port < 1 || port > 65535) {
+    return false;
+  }
+
+  return r->set_port(port_input);
```
anonrig, line 305: This is wrong. You should not update C API at all. C API calls url_aggregator, so if you update url_aggregator, this will work.

mertcanaltin: thanks I'll fix it now :pray:

tests/basic_tests.cpp
```diff
@@ -17,6 +17,7 @@ TYPED_TEST(basic_tests, insane_url) {
   ASSERT_EQ(r->get_hostname(), "");
   ASSERT_EQ(r->get_port(), "");
   ASSERT_EQ(r->get_pathname(), "@EEEEEEEEEE");
```
anonrig, line 19: This will help you catch more bugs:

```suggestion
  ASSERT_EQ(r->get_pathname(), "@EEEEEEEEEE");
  ASSERT_TRUE(r->is_valid);
```

tests/basic_tests.cpp
```diff
@@ -17,6 +17,8 @@ TYPED_TEST(basic_tests, insane_url) {
...
+  ASSERT_TRUE(r->is_valid);
+  ASSERT_FALSE(r->set_port("invalid80"));
```
anonrig, line 21: The order is important

```suggestion
  ASSERT_FALSE(r->set_port("invalid80"));
  ASSERT_TRUE(r->is_valid);
```

tests/basic_tests.cpp
mertcanaltin: I view this log: https://ctxt.io/2/AAB4CV4WFw

I put it on this site because it is too long

PR #836, opened 2025-01-10
Fix set_port to return correct boolean value

CarlosEduR:
Fix: https://github.com/ada-url/ada/issues/826


CarlosEduR:
It looks like checking that in the set_port (before the parse_port) does not cause some tests to fail... Now I am wondering about the fuzzers.

anonrig:
> It looks like checking that in the set_port (before the parse_port) does not cause some tests to fail... Now I am wondering about the fuzzers.

I think fuzzers are unrelated. It's failing on another PR.

src/url.cpp
```diff
@@ -791,6 +791,12 @@ bool url::set_port(const std::string_view input) {
     return false;
   }
 
+  uint16_t parsed_port{};
+  std::from_chars(trimmed.data(), trimmed.data() + trimmed.size(), parsed_port);
```
anonrig, line 795: This is going to be slower. can we do it inside parse_port()

CarlosEduR: that was my initial attempt here https://github.com/ada-url/ada/pull/836/commits/e998da70e4f6964689d94bdcea21d2b37c069e5e, but these tests started failing:
```
6 - wpt_url_tests_typed.urltestdata_encoding<ada::url> (Failed)
9 - wpt_url_tests_typed.urltestdata_encoding<ada::url_aggregator> (Failed)
16 - url_components.urltestdata_encoding (Failed)
```

image

was my approach wrong?

anonrig: I think "bad expected access" means that there is another bug that's surfaced because of your change. Which is good.

CarlosEduR: Got it, so I'll bring those changes back, and check if I can get more info about this bug.

CarlosEduR:
Closing in favor of https://github.com/ada-url/ada/pull/839

PR #850, opened 2025-01-24
update url_pattern to pass more wpt on node.js

anonrig:
This change is required to pass almost all WPT on Node.js

### DIFF
--- a/include/ada/url_pattern-inl.h
+++ b/include/ada/url_pattern-inl.h
@@ -47,11 +47,6 @@ url_pattern_component<regex_provider>::create_component_match_result(
   auto result =
       url_pattern_component_result{.input = std::string(input), .groups = {}};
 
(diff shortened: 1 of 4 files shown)


include/ada/url_pattern-inl.h
```diff
@@ -46,11 +46,6 @@ url_pattern_component<regex_provider>::create_component_match_result(
   auto result =
       url_pattern_component_result{.input = std::string(input), .groups = {}};
 
-  // If input is empty, then groups will always be empty.
```
anonrig, line 49: removing this breaks std::regex provider but succeeds V8 provider

jasnell (approved):

include/ada/url_pattern-inl.h
anonrig: We should move this to std_regex_provider...

PR #883, opened 2025-02-16
use gcc12 on s390x

anonrig:
Attempt to improve s390x build error ratio


.github/workflows/ubuntu-s390x.yml
```diff
@@ -34,9 +34,9 @@ jobs:
           githubToken: ${{ github.token }}
           install: |
             apt-get update -q -y
-            apt-get install -y cmake make g++ git ninja-build
+            apt-get install -y cmake make g++12 gcc12 git ninja-build
```
anonrig, line 37:
```suggestion
            apt-get install -y cmake make g++-12 gcc-12 git ninja-build
```

PR #887, opened 2025-02-17
reduce string allocations in url_pattern_init::process

anonrig:
Removes multiple string allocations and copies in url_pattern_init::process()


lemire (approved):
I love it.

PR #888, opened 2025-02-17
further reduce unnecessary string copies

anonrig:
Removes more string copies


anonrig:
@lemire @jasnell would you mind reviewing?

lemire (approved):

PR #891, opened 2025-02-24
enable -Wsign-compare to comply with node.js builds

anonrig:
let's enable this flag to catch regressions like https://github.com/nodejs/node/pull/57190


lemire (approved):
LGTM.

PR #901, opened 2025-03-01
fix: resolve std::system_error in adaparse on Windows

myd7349:
On Windows, the equivalent of `/dev/nul` is `nul`. Using `/dev/nul` on Windows will cause the program to throw a runtime exception when the output parameter is not specified.

### DIFF
--- a/tools/cli/adaparse.cpp
+++ b/tools/cli/adaparse.cpp
@@ -201,6 +201,12 @@ int piped_file(Callable&& adaparse_print, const cxxopts::ParseResult result,
 int main(int argc, char** argv) {
   std::ios::sync_with_stdio(false);
 
+#ifdef _WIN32
+  const char* devnull = "nul";
+#else
+  const char* devnull = "/dev/null";
+#endif
+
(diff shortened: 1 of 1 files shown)


anonrig (approved):
Nice. Is it easy to write a test for this?

myd7349:
Hello. @anonrig  Adding a test sounds like a good idea.  

To my understanding, this is a modification to `adaparse`, so should this test be added as a step in CI to execute `adaparse` and ensure it works(For example, checking whether it correctly parses a URL as expected or ensuring its exit code is 0?)? Or should a post-build event be added in CMake to run the test?  

I took a look at the tests under [https://github.com/ada-url/ada/tree/main/tests](https://github.com/ada-url/ada/tree/main/tests), and they all seem to be testing `ada`'s API. So, I need some guidance.

anonrig:
I think we need to add a new test file under tests folder that links to adaparse. Let's merge this as it is, and appreciate it if you can add a test for it as a follow up.

Issue #900, opened 2025-03-01
How to reduce library size?


ammarahm-ed:
I am looking for ways to reduce the size impact of ada which is around 500KB compiled by putting some features behind feature flags like Unicode support/URLPattern. There are environments like React Native/NativeScript where having a small binary size is crucial and many features are not needed. For example current RN polyfill for URL is this https://github.com/charpeni/react-native-url-polyfill which is 40.9KB in size after stripping out Unicode support. I am wondering if something similar is possible for ada. Thanks!

anonrig:
We can put URLSearchParams and URLPattern behind a compile flag so you don't add them to your build. But unfortunately, unicode support is required for URL parser to work correctly.

ammarahm-ed:
> We can put URLSearchParams and URLPattern behind a compile flag so you don't add them to your build. But unfortunately, unicode support is required for URL parser to work correctly. 

We do use URLSearchParams in NativeScript btw. However any size savings would be awesome. I am not sure how the JS polyfills work without Unicode included. Maybe they are not spec compliant or unable to handle complex urls.

anonrig:
> > We can put URLSearchParams and URLPattern behind a compile flag so you don't add them to your build. But unfortunately, unicode support is required for URL parser to work correctly.
> 
> We do use URLSearchParams in NativeScript btw. However any size savings would be awesome. I am not sure how the JS polyfills work without Unicode included. Maybe they are not spec compliant or unable to handle complex urls.

So, the only option is to make URLPattern optional. But even in that scenario, NativeScript uses URLPattern and can not be removed.

What I recommend is to actually pursue @lemire 's issue on ada/idna https://github.com/ada-url/idna/issues/54, which can save us around 37KB.

anonrig:
> Maybe they are not spec compliant or unable to handle complex urls.

@ammarahm-ed I opened an issue on their repository. https://github.com/charpeni/react-native-url-polyfill/issues/476 Their statement is wrong, and as far as I can see, their repository is 2 years old, which means their implementation does not follow the URL spec (by 2 years)

the-moisrex:
> What I recommend is to actually pursue [@lemire](https://github.com/lemire) 's issue on ada/idna [ada-url/idna#54](https://github.com/ada-url/idna/issues/54), which can save us around 37KB.

Actually it's about 110 KiB.

- IDNA Mapping tables: yours `84.47 KiB` -> mine `45.03 KiB`
[...]

PR #904, opened 2025-03-01
add an option to not include any urlpattern code

anonrig:
Partially fixes https://github.com/ada-url/ada/issues/900

cc @ammarahm-ed


ammarahm-ed:
That was quick! Thanks.

lemire (approved):

PR #909, opened 2025-03-08
apply new urlpattern tests

anonrig:
Depends on: https://github.com/web-platform-tests/wpt/pull/51220

cc @annevk @lemire @jasnell


lemire (approved):
This PR adds new tests, so it cannot cause much harm.

anonrig:
> This PR adds new tests, so it cannot cause much harm.

I've run WPT updater locally and added tests that pass with std::regex. Apparently some of them fails with Windows...

PR #957, opened 2025-06-13
fix: reject Windows file paths in can_parse

mertcanaltin:
reject Windows file paths in can_parse

I tried to solve this place out : https://github.com/nodejs/node/pull/58578#issuecomment-2970754561


src/implementation.cpp
```diff
@@ -49,6 +50,10 @@ std::string href_from_file(std::string_view input) {
 }
 
 bool can_parse(std::string_view input, const std::string_view* base_input) {
+  if (helpers::is_windows_file_path(input)) {
+    return false;
+  }
```
anonrig, line 55: Can you point us to where in the URL spec this is returning false? Chrome returns true...

```
> URL.canParse("file:///C:/path/file.node")
true
```

anonrig (changes requested):

src/implementation.cpp
anonrig: This is actually allowed. https://url.spec.whatwg.org/#file-host-state

If state override is not given and buffer is a [Windows drive letter](https://url.spec.whatwg.org/#windows-drive-letter), [file-invalid-Windows-drive-letter-host](https://url.spec.whatwg.org/#file-invalid-windows-drive-letter-host) [validation error](https://url.spec.whatwg.org/#validation-error), set state to [path state](https://url.spec.whatwg.org/#path-state).

anonrig: If you apply the same change to URL parser, not the can_parse() method, you'll see that it breaks web-platform tests.

mertcanaltin: thank you, instead of rejecting Windows file paths
Is it a healthy method to normalize them with the ```file:///``` protocol ?

lemire: What the library provides is WHATWG URL support. If the result disagrees with your expectations make sure first to check the WHATWG URL specification. If you disagree with it, you should take your concerns with WHATWG.

We may have bugs but a bug is a disagreement with WHATWG.

lemire:
At a glance this PR appears to break the specification.

mertcanaltin:
> At a glance this PR appears to break the specification.

@lemire Thank you for the clear guidance. I understand now that we should follow the WHATWG URL spec strictly. 

I've opened an issue in WHATWG to discuss the Windows file path handling: whatwg/url#873

I'll keep the library's behavior aligned with the spec while we discuss this in WHATWG.

lemire:
I am closing this PR since @mertcanaltin reported the issue to WHATWG. It is understood that if the standard changes, we will reopen such a PR.

PR #958, opened 2025-06-14
fix: optimize IPv4 hostname serialization

mertcanaltin:
I removed the duplicate serialization call and stored the result in a variable


anonrig (changes requested):
This doesn't change anything?

mertcanaltin:
> This doesn't change anything?

I wanted to try to keep it in a variable here, but I think the compiler already optimized it automatically. I will close this PR. thanks for the review!

PR #963, opened 2025-06-28
fix(url): handle :port values in host setter

watilde:
The current implementation incorrectly allows values like ":80" (empty host with port), resulting in `foo://:80/to`, while the spec requires failure if the host is missing. This update aligns behavior with the spec.

```js
const NodeURL = require("url").URL;
const WhatwgURL = require("whatwg-url").URL;
const newHost = ':80';

const url = new NodeURL("foo://path/to");  
url.host = newHost;
console.log(url.href); // Wrong: foo://:80/to

const url2 = new WhatwgURL("foo://path/to");
url2.host = newHost;
console.log(url2.href); // Correct: foo://path/to
```

Refs: https://url.spec.whatwg.org/#host-state
> If buffer is the empty string, host-missing validation error, return failure.

WPT tests for this case will be added in a follow-up PR.

### DIFF
--- a/src/url_aggregator.cpp
+++ b/src/url_aggregator.cpp
@@ -549,43 +549,75 @@ bool url_aggregator::set_host_or_hostname(const std::string_view input) {
     // Note: the 'found_colon' value is true if and only if a colon was
     // encountered while not inside brackets.
     if (found_colon) {
+      // If buffer is the empty string, host-missing validation error, return
+      // failure.
+      std::string_view host_buffer = host_view.substr(0, location);
+      if (host_buffer.empty()) {
+        return false;
+      }
+
+      // If state override is given and state override is hostname state, then
+      // return failure.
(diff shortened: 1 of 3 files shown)


src/url_aggregator.cpp
```diff
@@ -549,43 +549,75 @@ bool url_aggregator::set_host_or_hostname(const std::string_view input) {
...
+      }
+
+      bool succeeded = parse_host(host_view);
+      if (!succeeded) {
+        update_base_hostname(previous_host);
+        update_base_port(previous_port);
+        return false;
       } else if (has_dash_dot()) {
-        add_authority_slashes_if_needed();
+        // Should remove dash_dot from pathname
```
anonrig, line 616: What's the purpose of this?

lemire (approved):

anonrig (approved):

PR #982, opened 2025-07-22
Add missing includes of <iterator> header

dsanders11:
The `std::back_insert_iterator` function is defined in the `<iterator>` header, so it should be included in files which use that function.

### DIFF
--- a/src/url.cpp
+++ b/src/url.cpp
@@ -4,6 +4,7 @@
 
 #include <numeric>
 #include <algorithm>
+#include <iterator>
 #include <ranges>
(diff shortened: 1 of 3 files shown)


anonrig:
Thank you! Are you using Ada in any Microsoft projects?

dsanders11:
@anonrig, this came out of Electron's usage of Node.js. 🙂 The missing includes were causing a compilation error when we upgraded Chromium versions, I'm assuming `<iterator>` was previously being pulled in transitively allowing this to build for us before, but something changed.

PR #1000, opened 2025-09-24
add RVV support

camel-cdr:
This adds support for the RISC-V vector extension.

One the SpacemiT X60, I get speedups of 1.12x, 1.15x and 1.24x, for AdaURL_href, AdaURL_aggregator_href and AdaURL_CanParse respectively. This is in-line with, even above, the relative speedup from the SSE code paths when toggled on Zen1 (1.04x, 1.01x and 1.11x).

I decided to use LMUL=1 because URLs usually are on the smaller side.
There is a `#if`ed out codepath that implements `find_next_host_delimiter_special()` with a LUT, which was slower than the regular implementation on the X60, but I could see if being faster on future OoO hardware.

I didn't add runtime ISA detection yet, because I expect toolchains to move the default to RVA23 in the future. That would probably be better in a seperate PR anyways, as ada currently doesn't have the infrastructure for runtime dispatch.

### DIFF
--- a/cmake/toolchains-dev/riscv64-rvv.cmake
+++ b/cmake/toolchains-dev/riscv64-rvv.cmake
@@ -1,0 +1,19 @@
+# Usage:
+# $ cmake -DCMAKE_TOOLCHAIN_FILE=cmake/toolchains-dev/riscv64-rvv.cmake
+set(CMAKE_SYSTEM_NAME Generic)
+
+set(target       riscv64-linux-gnu)
+set(c_compiler   gcc)
+set(cxx_compiler g++)
+
+set(CMAKE_C_COMPILER   "${target}-${c_compiler}")
+set(CMAKE_CXX_COMPILER "${target}-${cxx_compiler}")
+
+set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
+
+set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
+
+set(CMAKE_CROSSCOMPILING_EMULATOR "qemu-riscv64")
+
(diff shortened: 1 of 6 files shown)


cmake/toolchains-dev/riscv64-rvv.cmake
```diff
@@ -0,0 +1,19 @@
+# Usage:
```
anonrig, line 1: Can you update/add necessary sections to cmake/toolchains-dev/readme.md?

include/ada/common_defs.h
```diff
@@ -247,6 +247,10 @@ namespace ada {
 #define ADA_LSX 1
 #endif
 
+#if defined(__riscv_v) && __riscv_v_intrinsic >= 11000
```
anonrig, line 250: Can you add a comment to why we have >= 11000?

src/helpers.cpp
```diff
@@ -355,6 +355,42 @@ ada_really_inline size_t find_next_host_delimiter_special(
   }
   return size_t(view.length());
 }
+#elif ADA_RVV
+ada_really_inline size_t find_next_host_delimiter_special(
+    std::string_view view, size_t location) noexcept {
+  // The LUT approach was a bit slower on the SpacemiT X60, but I could see it
+  // beeing faster on future hardware.
+#if 0
```
anonrig, line 363: if we're going to land this, we should remove all the unused parts

anonrig (commented):
I'll leave the actual review to @lemire

camel-cdr:
@dzaima I was thinking about that, but now the instructions are serialized and I'm not sure if this requires extra moves, or if you can have the destination and mask both be v0.

@anonrig I'll do the comments in a few.

I wanted to keep the unused code in so it's easier to test which is better once OoO RVV implementations are available. But I can also remove it.

PR #1020, opened 2025-11-15
add SSSE3 SIMD instructions

anonrig:
It seems SSSE3 gives us a 5% performance boost.

  1. BasicBench_AdaURL_href (Full URL parsing with href getter)

  | Metric               | SSE2 Baseline (median) | SSSE3 Optimized (median) | Improvement |
  |----------------------|------------------------|--------------------------|-------------|
  | Time per URL         | 249.276 ns             | 237.648 ns               | 4.7% faster |
  | Throughput           | 327.212 M/s            | 354.032 M/s              | 8.2% higher |
  | URLs per second      | 3.767M                 | 4.076M                   | 8.2% more   |
  | Time per byte        | 3.056 ns               | 2.825 ns                 | 7.6% faster |
  | Cycles per URL       | 974.434                | 964.386                  | 1.0% fewer  |
  | Instructions per URL | 3.218k                 | 3.190k                   | 0.9% fewer  |

  2. BasicBench_AdaURL_aggregator_href (Aggregator parsing)

  | Metric               | SSE2 Baseline (median) | SSSE3 Optimized (median) | Improvement |
  |----------------------|------------------------|--------------------------|-------------|
  | Time per URL         | 155.386 ns             | 150.837 ns               | 2.9% faster |
  | Throughput           | 517.071 M/s            | 526.278 M/s              | 1.8% higher |
  | URLs per second      | 5.953M                 | 6.059M                   | 1.8% more   |
  | Time per byte        | 1.934 ns               | 1.900 ns                 | 1.8% faster |
  | Cycles per URL       | 619.107                | 603.857                  | 2.5% fewer  |
  | Instructions per URL | 2.209k                 | 2.178k                   | 1.4% fewer  |

### DIFF
--- a/include/ada/common_defs.h
+++ b/include/ada/common_defs.h
@@ -233,6 +233,10 @@ namespace ada {
   } while (0)
 #endif
 
+#if defined(__SSSE3__)
+#define ADA_SSSE3 1
+#endif
+
 #if defined(__SSE2__) || defined(__x86_64__) || defined(__x86_64) || \
     (defined(_M_AMD64) || defined(_M_X64) ||                         \
      (defined(_M_IX86_FP) && _M_IX86_FP == 2))
--- a/src/helpers.cpp
+++ b/src/helpers.cpp
@@ -5,6 +5,10 @@
 #include "ada/common_defs.h"
 #include "ada/scheme.h"
 
+#if ADA_SSSE3
+#include <tmmintrin.h>
+#endif
+
 namespace ada::helpers {
 
 template <typename out_iter>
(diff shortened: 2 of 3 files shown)


lemire (approved):
This is reasonable, but note that unless the library is compiled for SSSE3 support, the new path will not be used. And ada will typically not be compiled for SSSE3 (unless you are a gentoo user or something of the sort), thus the support will not be used. (Note that whether the cpu defaults on SSSE3 support depends on the system. So there are definitely systems where your optimization will work out of the box. I think Suse and RedHat maybe?)

One possibility, if we don't want to manually do the runtime dispatching (like in simdutf and simdjson) would be to specialize the feature for GCC/clang. (So no Visual Studio support.)

Then you can use `__builtin_cpu_supports`.

https://clang.llvm.org/docs/LanguageExtensions.html

https://gcc.gnu.org/onlinedocs/gcc/x86-Built-in-Functions.html

It is likely that copilot/claude/... can do it for you pretty close to correctly.

One downside of such an optimization is that it makes testing more difficult because you have to test two or more paths depending on what the CPU supports. In a library like simdjson or simdutf, we do everything manually.

PR #1032, opened 2025-12-08
URLPattern: replace std::string with std::string_view

CarlosEduR:
I noticed that WebKit already uses a StringView for these parameters.


GuillaumeLagrange:
Google benchmark has features for this kind of usage

https://github.com/google/benchmark/blob/main/docs/user_guide.md#templated-benchmarks-that-take-arguments

If you have any question/issues about how each benchmark is identified with codspeed's google benchmark layer, do not hesitate to create a dedicated issue on https://github.com/CodSpeedHQ/codspeed-cpp
You should also be able to manually name the benchmarks.

PR #1037, opened 2025-12-10
Make some ifs constexpr

the-moisrex:
This makes some `if`s `if constexpr`s even though it has no affect in the benchmarks, but it's better to be explicit.

Also removed a useless `enum`.

### DIFF
--- a/benchmarks/benchmark_template.cpp
+++ b/benchmarks/benchmark_template.cpp
@@ -14,10 +14,7 @@ size_t count_ada_invalid() {
   return how_many;
 }
 
-enum { JUST_PARSE = 1, PARSE_AND_HREF = 0 };
-
-template <bool just_parse = PARSE_AND_HREF,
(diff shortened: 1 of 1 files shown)


the-moisrex:
Also, I'm getting these warnings:

```
/ada-url/benchmarks/benchmark_template.cpp:27:9: warning: ‘++’ expression of ‘volatile’-qualified type is deprecated [-Wvolatile]
   27 |         success++;
      |         ^~~~~~~
/ada-url/benchmarks/benchmark_template.cpp:42:11: warning: ‘++’ expression of ‘volatile’-qualified type is deprecated [-Wvolatile]
   42 |           success++;
      |           ^~~~~~~
/ada-url/benchmarks/benchmark_template.cpp: In function ‘void Bench_BasicBench_AdaURL_CanParse(benchmark::State&)’:

```

anonrig (approved):

Issue #1061, opened 2026-01-18
Minor: Silence CodeQL warning in has_non_empty_password()


nzrsky:
Hi!

I'm using ada in my project and noticed that CodeQL flags this line:

```cpp
// url_aggregator-inl.h:777
return components.host_start - components.username_end > 0;
```

The warning is `cpp/unsigned-difference-expression-compared-zero` (CWE-191).

Of course, I realize that `host_start < username_end` can never actually happen given URL structure invariants — so this isn't a real issue.

Would you consider changing it to:

```cpp
return components.host_start > components.username_end;
```

Thanks for the great library!

nzrsky:
Sure, done.

anonrig:
Thank you

PR #1062, opened 2026-01-18
resolve cpp/unsigned-difference-expression-compared-zero warning (#1061)

nzrsky:
This PR resolves the cpp/unsigned-difference-expression-compared-zero warning in the URL aggregator implementation (https://github.com/ada-url/ada/issues/1061)


lemire (approved):
This is good.

PR #1079, opened 2026-02-23
fix urlpattern OOB read

anonrig:


jasnell (approved):

include/ada/url_pattern-inl.h
```diff
@@ -39,13 +39,14 @@ url_pattern_component<regex_provider>::create_component_match_result(
...
   // We explicitly start iterating from 0 even though the spec
   // says we should start from 1. This case is handled by the
-  // std_regex_provider.
-  for (size_t index = 0; index < exec_result.size(); index++) {
+  // std_regex_provider which removes the full match from index 0.
+  // Use min() to guard against potential mismatches between
+  // exec_result size and group_name_list size.
+  const size_t size = std::min(exec_result.size(), group_name_list.size());
+  result.groups.reserve(size);
+  for (size_t index = 0; index < size; index++) {
```
lemire: As far as I can tell, this fix only fixes something if we have a case where `exec_result.size()` is greater than `group_name_list.size()`. Running our tests with such a check, I do not find any case. Did you reproduce an OOB?

Issue #1082, opened 2026-02-27
Add support for building against system simdutf


kapouer:
### What is the problem this feature will solve?

Cmake os doing weird stuff, when I just want to compile with system-installed simdutf debian package.
Configure flags are:
 -DADA_TOOLS=False -DBUILD_SHARED_LIBS=True -DCMAKE_DISABLE_FIND_PACKAGE_Python3=True -DADA_USE_SIMDUTF=ON -DADA_LIB_ABI=0

### What is the feature you are proposing to solve the problem?

I have no idea. A cmake flag ?

### What alternatives have you considered?

I need to do this i nCMakeLists.txt:

```
  if(ADA_USE_SIMDUTF)
    #CPMAddPackage(
    #  NAME simdutf
    #  GITHUB_REPOSITORY simdutf/simdutf
    #  VERSION 7.3.2
    #  OPTIONS "SIMDUTF_TESTS OFF" "SIMDUTF_TOOLS OFF"
    #)
  endif()
```

kapouer:
@lemire thank you.
I panicked with CPM and didn't do my homework.

PR #1083, opened 2026-02-27
add an option to use shared simdutf

anonrig:
Fixes https://github.com/ada-url/ada/issues/1082


kapouer:
Cool ! I'm going to try it.

PR #1089, opened 2026-03-18
Improving the perf of the scheme matching

lemire:
Make is so that `get_scheme_type` and `get_special_port` do not trigger a mispredicted branch when the scheme is a special scheme (and most schemes are special, so...). We also reduce the instruction count.

This is not a large enough fraction of our standard URL parsing routines, so I created a new microbenchmark (`bench_protocol`) specifically to test it out.

Here is what you might get on an Apple platform:

```
legacy                                             :  4.647 ns   0.22 Gv/s   4.40 GHz  20.99 c  57.45 i   2.74 i/c   0.57 bm 
hand-tuned hash                                    :  0.948 ns   1.05 Gv/s   4.31 GHz   4.32 c  41.57 i   9.62 i/c   0.00 bm 
ada                                                :  1.018 ns   0.98 Gv/s   4.27 GHz   4.68 c  43.82 i   9.37 i/c   0.00 bm 
```

So fewer instructions *and* fewer mispredicted branches !!! Win, win.

### DIFF
--- a/include/ada/scheme-inl.h
+++ b/include/ada/scheme-inl.h
@@ -21,6 +21,7 @@ constexpr std::string_view is_special_list[] = {"http", " ",   "https", "ws",
 // for use with get_special_port
 constexpr uint16_t special_ports[] = {80, 0, 443, 80, 21, 443, 0, 0};
 
+// @private
 // convert a string_view to a 64-bit integer key for fast comparison
 constexpr uint64_t make_key(std::string_view sv) {
   uint64_t val = 0;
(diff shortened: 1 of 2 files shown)


include/ada/scheme-inl.h
```diff
@@ -20,6 +20,37 @@ constexpr std::string_view is_special_list[] = {"http", " ",   "https", "ws",
                                                 "ftp",  "wss", "file",  " "};
 // for use with get_special_port
 constexpr uint16_t special_ports[] = {80, 0, 443, 80, 21, 443, 0, 0};
+
+// convert a string_view to a 64-bit integer key for fast comparison
```
anonrig:
```suggestion
// @private
// convert a string_view to a 64-bit integer key for fast comparison
```

PR #1091, opened 2026-03-20
Clarify borrowed ada_string lifetime and add owned host accessor

rootvector2:
Fixes a lifetime issue in the C API where `ada_string` returned from getters is a borrowed view into internal storage and becomes invalid after mutation.

- Documented lifetime behavior of `ada_string` in `ada_c.h`
- Added `ada_get_host_owned` to provide a safe, owned copy
- Added a small test to demonstrate correct usage

No breaking changes.

### DIFF
--- a/include/ada_c.h
+++ b/include/ada_c.h
@@ -14,6 +14,12 @@
 #define ada_url_omitted 0xffffffff
 
 // string that is owned by the ada_url instance
+//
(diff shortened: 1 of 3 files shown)


anonrig (approved):

anonrig:
@lemire would you mind taking a look?

PR #1121, opened 2026-04-05
perf: avoid heap allocations on URL parsing failure

anonrig:
Two complementary optimizations targeting the parse failure path:

1. Defer url_aggregator::reserve() to after scheme parsing
   Previously the buffer capacity was reserved upfront (power-of-two
   rounded from input length) before the state machine ran.  For inputs
   that never reach a valid scheme -- bare paths with no base, strings
   starting with a digit, etc. -- this caused an unconditional heap
   allocation that was immediately thrown away.

   The reserve() call is now deferred to inside the SCHEME state, right
   after parse_scheme_with_colon() writes the scheme bytes.  At that
   point the scheme is already in the buffer (at most ~8 bytes, within
   SSO on every modern implementation), so the subsequent reserve()
   triggers at most one heap allocation plus a small SSO-to-heap copy.
   The success-path allocation count is unchanged.

2. Release heap memory on failure via return_invalid() helper
   When parsing fails mid-way (invalid host, port out of range, empty
   authority after '@', etc.) the URL object was returned carrying a
   partially-filled buffer (url_aggregator) or a set of individually
   allocated strings (url), even though the caller only needs to observe
   is_valid == false and then discard the object.

   A noexcept lambda return_invalid() is introduced that:
   - For url_aggregator: swaps the buffer with an empty string (freeing
     the reserved capacity) and resets components to their default state.
   - For url: releases non_special_scheme, username, password, path, and
     resets host/query/hash/port optionals.

   All failure-path early returns inside the state machine now go through
   this helper instead of returning the partially-constructed url directly.

https://claude.ai/code/session_01LE4dy9buzBJ1cJ83wxKk58

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # common build directory
 build
 build-*/
+build_*/
 *-build-*
 _fuzz_check/
 
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -67,10 +67,9 @@ result_type parse_url_impl(std::string_view user_input,
     // This rounds up to the next power of two.
     // We know that user_input.size() is in [0,
     // std::numeric_limits<uint32_t>::max).
-    reserve_capacity =
-        (0xFFFFFFFF >>
-         helpers::leading_zeroes(uint32_t(1 | user_input.size()))) +
-        1;
+    reserve_capacity = (0xFFFFFFFF >> helpers::leading_zeroes(
+                                          uint32_t(1 | user_input.size()))) +
+                       1;
     // Note: url.reserve(reserve_capacity) is called in the SCHEME state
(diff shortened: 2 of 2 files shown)


anonrig:
Claude doesn't realize how our benchmarks work...

lemire (approved):
I don't mind, this being said, if you are spending a lot of time parsing invalid URLs, one has other problems than heap allocation.

anonrig:
> I don't mind, this being said, if you are spending a lot of time parsing invalid URLs, one has other problems than heap allocation.

I don't see any improvement on any benchmark. I think I'll just close this... Adds unnecessary complexity for almost no benefit (except scheme validation case)

PR #1123, opened 2026-04-06
Harden URL component offset updates by eliminating unsigned wraparound

metsw24-max:
This patch replaces unsigned wraparound-based arithmetic used in URL component offset updates with checked explicit arithmetic

Adds centralized helpers `checked_shift_offset`, `compute_shifted_non_scheme_offsets`, `apply_shifted_non_scheme_offsets`

Adds safe size_t → uint32_t conversion checks

Ensures fail-closed behavior `is_valid = false` on invalid transformations

Removes reliance on implicit unsigned overflow in offset calculations

### DIFF
--- a/src/url_aggregator.cpp
+++ b/src/url_aggregator.cpp
@@ -9,9 +9,104 @@
 #include "ada/url_aggregator-inl.h"
 
 #include <iterator>
+#include <limits>
+#include <optional>
 #include <ranges>
 #include <string>
 #include <string_view>
+
+namespace {
+
+constexpr uint32_t kMaxUrlComponentOffset =
+    std::numeric_limits<uint32_t>::max();
+
+struct shifted_non_scheme_offsets {
+  uint32_t username_end;
+  uint32_t host_start;
+  uint32_t host_end;
+  uint32_t pathname_start;
+  uint32_t search_start;
+  uint32_t hash_start;
+};
+
+[[nodiscard]] ada_really_inline bool checked_shift_offset(uint32_t offset,
+                                                          int64_t delta,
+                                                          uint32_t& out) {
+  if (delta >= 0) {
+    const uint64_t positive_delta = static_cast<uint64_t>(delta);
+    if (positive_delta >
+        static_cast<uint64_t>(kMaxUrlComponentOffset - offset)) {
+      return false;
+    }
+    out = offset + static_cast<uint32_t>(positive_delta);
+    return true;
+  }
+
+  const uint64_t negative_delta = static_cast<uint64_t>(-delta);
+  if (negative_delta > static_cast<uint64_t>(offset)) {
+    return false;
+  }
+  out = offset - static_cast<uint32_t>(negative_delta);
+  return true;
+}
+
+[[nodiscard]] std::optional<shifted_non_scheme_offsets>
+compute_shifted_non_scheme_offsets(const ada::url_components& components,
+                                   int64_t delta) {
+  shifted_non_scheme_offsets shifted{components.username_end,
+                                     components.host_start,
+                                     components.host_end,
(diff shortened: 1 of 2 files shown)


lemire:
@metsw24-max Is an overflow ever possible given that we have a hard limit on the size of the URLs?

metsw24-max:
You are right that the parser enforces a size limit so in normal cases that already keeps values well below the `uint32_t` boundary

The intent here is to avoid relying on unsigned wraparound altogether 

The previous code assumed overflow was acceptable because it is well defined, but that can still lead to subtle state corruption when offsets are shifted repeatedly

With explicit checks the offset updates either succeed safely or fail cleanly which makes the logic easier to reason about and more robust overall

lemire:
@metsw24-max 

> that can still lead to subtle state corruption when offsets are shifted repeatedly

What does that mean?

anonrig:
What's the motivation for this change?

lemire:
For context, most programming languages (the only exception I know is Swift) do not check for the overflows in release mode. In C/C++, you can check for overflows with `-fsanitize=unsigned-integer-overflow` (sometimes enabled in debug mode), Rust does it in debug mode. C# does not care unless you put the code within a `checked` scope. And so forth. 

I don't think it is warranted here. *If it happens* then we have a nasty bug. Returning false would not save us. So if we think it can happen, we should find out. I don't think it can happen... simply because you are not going to be parsing strings that are gigabytes... 

I do like the `apply_shifted_non_scheme_offsets` but not because of the bound checking, but because having a dedicated function could trim some redundancy.

metsw24-max:
I agree that overflow is not really expected in practice given the URL size constraints and current invariants.

My earlier wording around “subtle state corruption” could have been clearer......what I meant is that the previous code relies on unsigned wraparound to express shifts, which works, but makes the intent a bit harder to follow and reason about locally.

The main goal of this change was to centralize and deduplicate the offset shifting logic using `apply_shifted_non_scheme_offsets` and make those updates clearer and easier to maintain.

The overflow checks were more of a defensive addition......so I can simplify this to focus just on the refactor by keeping `apply_shifted_non_scheme_offsets` and removing the additional bounds-checking logic so the behavior stays unchanged while reducing duplication.

Does that direction sound good?

bbayles:
You can elicit some strange behavior by exploiting the wraparound. Illustrated using the Python wrapper:

```python3
>>> from ada_url import URL
>>> u = URL('https://example.org')
>>> u.pathname = '?' * 1431655764
>>> len(u.pathname)
4294967293
>>> u.pathname = '?' * 1431655765
>>> len(u.pathname)
0
>>> u.pathname
''
>>> u.href[:40]
'https://example.org/%3F%3F%3F%3F%3F%3F%3'
```

lemire:
@bbayles Please see https://github.com/ada-url/ada/pull/1126

metsw24-max:
@lemire 
I see how this ties into #1126 and the approach of enforcing stricter size limits to prevent these edge cases up front rather than handling them through local checks.

I will keep this PR focused on the refactor and simplify it by retaining `apply_shifted_non_scheme_offsets` for deduplication and removing the additional bounds-checking logic so the behavior stays unchanged.

Does that direction sound good?

lemire:
@metsw24-max I think `apply_shifted_non_scheme_offsets` is nice.

metsw24-max:
@lemire 
I kept only the dedup refactor by retaining `apply_shifted_non_scheme_offsets` and wiring it into the existing scheme offset update paths.
I removed the additional bounds-checking and fail-closed logic, and removed the related hardening tests, so behavior is unchanged and the PR stays narrowly scoped to refactoring

lemire (approved):

PR #1126, opened 2026-04-08
Enforce max input length to the Ada URL parser

lemire:
Up until now, we limited the size of the input string when parsing (to 4 GB). This could not be changed (it was hardcoded). Further, it was weakly enforced. It was possible to receive a string that was under 4GB and produce a larger normalized string. Further, the setters would enforce no length limit.

That's not much of a weakness because if you have a 4 GB URL, you have other problems.

BUT: we can cheaply set a different default. Say: you want all URLs to fit in 64kB. If the normalized URL exceeds 64kB, then you fail, that's it.

To make this efficient, we need a `get_href_size()` to save on some of the checks.

AI is telling me that once you get to 8kB or beyond, URLs become unusable in practice due to the limits set at the server level. I am not arguing we set a hard limit, but many users would definitely want to put a reasonable limit as part of their system's design. I cannot see any reason for a URL to exceed 2 MB in the real world. It is almost certainly a bug or an attack.

The fun thing is that we can now fuzz this easily. Set a short limit (like 512 bytes) and you can test it out and make sure that you fail before you exceed the limit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,14 +183,14 @@ url->set_hash("is-this-the-real-life"); // Update hash/fragment
 
 By default, ada allows URLs up to about 4 GB. You can set a lower limit to
 reject any URL whose serialized form (the href) would exceed a given number
-of bytes. The limit is enforced during parsing and across all setters. If
-a setter would cause the URL to exceed the limit, the call fails and the URL
-is left unchanged. Percent-encoding expansion is accounted for: a short input
-that encodes into a long result is still rejected.
+of bytes. The limit is enforced during parsing and across all setters.
+Setters that return `bool` return `false` when the limit would be exceeded;
+`void` setters (`set_search`, `set_hash`) silently leave the URL unchanged.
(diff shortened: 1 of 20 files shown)


anonrig (approved):

anonrig:
clang-tidy seems to fail

lemire:
> clang-tidy seems to fail

False positive. It cannot happen. The tool is not smart enough.

lemire:
@anonrig Next release should be a major release.

bbayles:
For the Python bindings, I'm inclined to set a maximum of 256 KiB and document that it's possible to change it at the process level. Seems like a generous default to me, but if we're aware of applications that need more, I could choose something larger.

lemire:
@bbayles gunicorn has different limits  by default...

https://gunicorn.org/reference/settings/

For basic http, it is going to be 4kB. It seems that 256 KiB  is plenty.

PR #1136, opened 2026-05-04
incomplete IDNA Bidi validation allowing invalid LTR labels with disallowed trailing code points

jmestwa-coder:
This change fixes incomplete IDNA Bidi validation for LTR labels where the last non-NSM code point was not being validated.

The issue was caused by the loop stopping before last_non_nsm_char, which skipped validation of the final non-NSM character. As a result, labels ending with disallowed Bidi classes were incorrectly accepted.

For example:

a\u05D0 (ends with RTL)
a\u0627 (ends with AL)
a\u0661 (ends with AN)

This violates RFC 5893 rules requiring that all characters in LTR labels must belong to allowed classes and that the final non-NSM character must be L or EN.

The fix extends validation to include the last non-NSM code point and explicitly enforces that it must be L or EN. No other logic is changed.

Regression tests have been added to cover:

LTR labels ending in RTL, AL, and AN
Multiple trailing RTL characters
NSM interactions

All new test cases fail before the fix and pass after it. The full test suite was run and no regressions were observed.

The change is limited to LTR Bidi validation logic in ada_idna.cpp with corresponding test additions in toascii.json.

### DIFF
--- a/src/ada_idna.cpp
+++ b/src/ada_idna.cpp
@@ -9407,18 +9407,19 @@ bool is_label_valid(const std::u32string_view label) {
 
       // In an LTR label, only characters with the Bidi properties L, EN,
       // ES, CS, ET, ON, BN, or NSM are allowed.
-      for (size_t i = 0; i < last_non_nsm_char; i++) {
+      for (size_t i = 0; i <= last_non_nsm_char; i++) {
         const direction d = find_direction(label[i]);
         if (!(d == direction::L || d == direction::EN || d == direction::ES ||
               d == direction::CS || d == direction::ET || d == direction::ON ||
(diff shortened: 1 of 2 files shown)


anonrig:
Can you open a PR to github.com/ada-url/idna as well?

jmestwa-coder:
@anonrig 
I verified the same issue exists in ada-url/idna and opened a corresponding PR here:

https://github.com/ada-url/idna/pull/74

It applies the same minimal fix and regression tests.

anonrig (approved):

lemire:
@jmestwa-coder Thanks for the report and fix.

PR #1138, opened 2026-05-11
Harden IDNA post-decode label validation

metsw24-max:
Harden the IDNA ToUnicode pipeline by validating decoded punycode labels before emitting Unicode output.

Previously, labels produced by the punycode decoder were emitted directly after decoding. This patch adds defensive post-decode validation to ensure decoded labels preserve expected IDNA normalization and validity invariants before conversion to UTF-8.

If validation fails the original ACE/punycode label is preserved instead of emitting a potentially malformed or non-canonical Unicode label.

## Changes

* add post-decode validation for punycode labels
* reject ASCII-only decoded labels
* verify mapping stability after decode
* verify normalization stability after decode
* validate decoded labels before UTF-8 emission
* preserve safe fallback behavior by returning the original ACE label when validation fails
* add regression coverage for malformed decoded punycode labels

## Security Impact

This improves the safety and auditability of the IDNA canonicalization pipeline by preventing malformed or non-canonical decoded labels from being emitted as trusted Unicode output.

The change reduces ambiguity in hostname canonicalization and strengthens invariant enforcement in security-sensitive URL and IDNA processing paths.

### DIFF
--- a/src/ada_idna.cpp
+++ b/src/ada_idna.cpp
@@ -9727,6 +9727,7 @@ std::string to_unicode(std::string_view input) {
 
   size_t label_start = 0;
   std::u32string tmp_buffer;
+  std::u32string post_map;
   while (label_start < input.size()) {
     size_t loc_dot = input.find('.', label_start);
     bool is_last_label = (loc_dot == std::string_view::npos);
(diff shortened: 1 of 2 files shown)


anonrig:
This needs to be first opened to ada-url/idna repository.

metsw24-max:
> This needs to be first opened to ada-url/idna repository.

i have raised in [ada-url/idna#75](https://github.com/ada-url/idna/pull/75)

lemire:
@metsw24-max Ok, but let us do things in order.

When idna is updated, we push the updates directly to ada and then update.

idna => ada 

So I am going to close this issue here.

There is no need to duplicate the issues.

PR #1139, opened 2026-05-11
Fix set_protocol slow-path return consistency

metsw24-max:
Fix inconsistent return values in the slow-path state-override handling for:

- `url::parse_scheme<true>`
- `url_aggregator::parse_scheme_with_colon<true>`

Several WHATWG validation failures correctly rejected the protocol change and left the URL unchanged, but incorrectly returned `true` when the target scheme was non-special.

The equivalent fast-path logic already returned `false` for the same rejected mutations, resulting in inconsistent API behavior.

This patch aligns both paths with the documented setter contract:
- return `true` on successful mutation
- return `false` when the scheme change is rejected

## Tests

Adds coverage for rejected non-special target scheme mutations and verifies:
- rejected mutations return `false`
- URL state remains unchanged
- valid non-special to non-special transitions still succeed

## Impact

Behavioral consistency fix only.
No changes to successful parsing or mutation behavior.

### DIFF
--- a/src/url_aggregator.cpp
+++ b/src/url_aggregator.cpp
@@ -90,25 +90,22 @@ template <bool has_state_override>
     unicode::to_lower_ascii(_buffer.data(), _buffer.size());
 
     if constexpr (has_state_override) {
-      // If url's scheme is a special scheme and buffer is not a special scheme,
-      // then return. If url's scheme is not a special scheme and buffer is a
-      // special scheme, then return.
+      // The state-override validation errors below ("return" in the WHATWG URL
+      // parser) leave the URL unchanged. The setter contract is
+      // "true on success, false if the scheme is invalid" -- the fast path
+      // above already returns false here, so the slow path must agree.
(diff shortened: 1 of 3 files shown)


src/url_aggregator.cpp
```diff
@@ -90,25 +90,22 @@ template <bool has_state_override>
...
+      // parser) leave the URL unchanged. The setter contract is
+      // "true on success, false if the scheme is invalid" -- the fast path
+      // above already returns false here, so the slow path must agree.
       if (is_special() != ada::scheme::is_special(_buffer)) {
-        return true;
+        return false;
       }
 
-      // If url includes credentials or has a non-null port, and buffer is
-      // "file", then return.
```
anonrig, line 101: Can you read these comments? They're taken from the spec and it helps us keep track of the changes.

metsw24-max: restrored it

anonrig (approved):
Thank you

PR #1141, opened 2026-05-11
updating idna to 0.5.2

lemire:
Self-describing.


anonrig (approved):

anonrig:
Tests are failing

lemire:
@anonrig Someone used AI and it injected non-ASCII chars.

Easy fix.

lemire:
Problem is with idna, upstream https://github.com/ada-url/idna/pull/76

AI wants to assume UTF-8 but we want to support cases where people do not default on UTF-8 (common under Windows).

lemire:
Hopefully, it will pass now.

lemire:
@anonrig This is good now.

PR #1146, opened 2026-05-25
fix out-of-bounds read in valid_name_code_point

metsw24-max:
Found while fuzzing the URLPattern constructor parser. valid_name_code_point's non-first branch runs std::lower_bound over id_continue but then checks the result against std::end(id_start), a different array. A name code point past the last id_continue range (above U+E01EF, e.g. U+10FFFF in a `:name`) makes lower_bound return end(id_continue), so the wrong end check passes and `(*iter)[0]` reads one element past id_continue.


anonrig (approved):

anonrig:
Can you also open the same PR to ada/idna?

lemire:
Reported against the wrong repo. Closing.

lemire:
See https://github.com/ada-url/idna/pull/79

PR #1156, opened 2026-06-22
fix: normalized Windows drive letter must be exactly two code points

spokodev:
`is_normalized_windows_drive_letter` accepted any input of length `>= 2` beginning with `<ASCII alpha>:`. Per the URL Standard a [normalized Windows drive letter](https://url.spec.whatwg.org/#normalized-windows-drive-letter) is a [Windows drive letter](https://url.spec.whatwg.org/#windows-drive-letter), which is two code points, so the check should require exactly two.

The loose length let [shorten a url's path](https://url.spec.whatwg.org/#shorten-a-urls-path) treat a longer first segment that merely starts with `<alpha>:` as a drive letter and skip popping it on `..`:

```
ada::parse("file:c:x/..")    -> "file:///c:x/"    (expected "file:///")
ada::parse("file:u:p@h/..")  -> "file:///u:p@h/"  (expected "file:///")
```

A real drive letter is unaffected: `file:c:/..` still yields `file:///c:/`.

The fix changes `>= 2` to `== 2`. All three call sites (`shorten_path` and the base-path drive-letter check in `parser.cpp`) pass a single path segment that should match only an exact drive letter, so the tighter check is correct for each.

Verified against the full WHATWG URL web-platform-tests `urltestdata` (no regressions) and added a regression test.

### DIFF
--- a/include/ada/checkers-inl.h
+++ b/include/ada/checkers-inl.h
@@ -47,7 +47,7 @@ constexpr bool is_windows_drive_letter(std::string_view input) noexcept {
 
 constexpr bool is_normalized_windows_drive_letter(
     std::string_view input) noexcept {
-  return input.size() >= 2 && (is_alpha(input[0]) && (input[1] == ':'));
+  return input.size() == 2 && (is_alpha(input[0]) && (input[1] == ':'));
 }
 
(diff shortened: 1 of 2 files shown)


lemire:
running tests

anonrig (approved):

PR #1157, opened 2026-06-22
fix: set empty host on a non-special URL without an authority

spokodev:
`url_aggregator::set_host` / `set_hostname` leaves the URL unchanged when an empty host is set on a non-special URL that has a non-opaque path but no authority (e.g. `foo:/bar`). `ada::url` already gives it an empty authority, so the two implementations disagree:

```cpp
auto a = ada::parse<ada::url>("non-special:/x");
a->set_host("");            // -> "non-special:///x"  (correct)

auto b = ada::parse<ada::url_aggregator>("non-special:/x");
b->set_host("");            // -> "non-special:/x"    (wrong, no-op)
```

Per the [host setter](https://url.spec.whatwg.org/#dom-url-host) steps the early return only happens for an opaque path. `non-special:/x` has a list path (`has_opaque_path()` is false), so an empty host must be set, which for a URL without an authority means adding one. `jsdom/whatwg-url` (the spec reference implementation, passing all wpt setter cases) also yields `non-special:///x`.

In `set_host_or_hostname` the empty-host / non-special branch handled the existing-hostname and dash-dot cases but otherwise fell through to a no-op. It now calls `add_authority_slashes_if_needed()`, matching `ada::url`.

Verified against the full WHATWG URL web-platform-tests `urltestdata` and `setters_tests` for both `url` and `url_aggregator` (no regressions), plus a typed regression test.

### DIFF
--- a/src/url_aggregator.cpp
+++ b/src/url_aggregator.cpp
@@ -677,6 +677,10 @@ bool url_aggregator::set_host_or_hostname(const std::string_view input) {
         } else if (has_dash_dot()) {
           add_authority_slashes_if_needed();
           delete_dash_dot();
+        } else {
+          // The url has no authority yet (e.g. "foo:/bar"); setting an empty
+          // host must still give it an (empty) authority, matching ada::url.
+          add_authority_slashes_if_needed();
         }
         return check_url_size();
       }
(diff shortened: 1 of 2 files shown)


anonrig (approved):

anonrig:
can you rebase and fix conflicts?

spokodev:
Rebased onto main and resolved the conflict. It was only in the tests: the drive-letter test from #1156 is upstream now, so I kept both. The fix itself is unchanged.

anonrig (approved):

PR #1167, opened 2026-07-03
fix canonicalize_hostname fast path skipping ipv4 and idna

Arawoof06:
canonicalize_hostname takes a fast path that returns the input unchanged when every byte is lowercase ascii, a digit, '-' or '.'. that set also matches ipv4-shaped hosts and xn-- labels, but the slow path it skips runs the value through a special-scheme dummy url which reserializes the address and validates idna, so the fast path leaves '0' as '0' instead of '0.0.0.0', '0x7f.1' instead of '127.0.0.1', and accepts an invalid ace label like 'xn--a' that set_hostname rejects. that makes a urlpattern built from a bare-number or punycode hostname disagree with its own canonical form. keep the fast path but fall through to the slow path when checkers::is_ipv4 is true or the value contains 'xn--'.

### DIFF
--- a/src/url_pattern_helpers.cpp
+++ b/src/url_pattern_helpers.cpp
@@ -9,6 +9,7 @@
 #include <string>
 
 #include "ada/character_sets.h"
+#include "ada/checkers.h"
 #include "ada/helpers.h"
 #include "ada/scheme.h"
 #include "ada/unicode.h"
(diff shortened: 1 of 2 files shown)


lemire (approved):
Looks like a reasonable fix for the url_pattern API. Note that this does not (as far as I can tell) affect the main URL parser.

PR #1168, opened 2026-07-04
normalize leading-zero ports in canonicalize_port

Arawoof06:
canonicalize_port in the URLPattern helpers validates the port with a hand-rolled length check and rejects any multi-digit run that starts with a zero. The URL parser reads a port as an integer though, so leading zeros belong to the value: `0080` is 80 and `065535` is 65535, both valid, matching what `new URL("https://x:0080").port` gives. As a result a URLPattern built with `{ port: "0080" }` throws while the equivalent URL parses fine, and the five-digit cap also miscounts because it includes the zeros. Strip leading zeros before the range check and return `0` for an all-zero value, the same way the fast path in implementation.cpp already does. canonicalize_port_with_protocol parses through from_chars and was already correct.

### DIFF
--- a/src/url_pattern_helpers.cpp
+++ b/src/url_pattern_helpers.cpp
@@ -391,8 +391,18 @@ tl::expected<std::string, errors> canonicalize_port(
   std::string_view digits_to_parse =
       std::string_view(trimmed.data(), first_non_digit - trimmed.begin());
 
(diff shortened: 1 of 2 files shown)


lemire (approved):
This looks credible.

anonrig (approved):

PR #1169, opened 2026-07-05
roll back set_host_or_hostname fully when host parse fails

Arawoof06:
set_host_or_hostname rolls back a failed host parse by calling update_base_hostname with the previous host, but that helper always runs add_authority_slashes_if_needed, so a rejected set_host or set_hostname on an authority-less non-special URL like `non-spec:/x` leaves a stray `//` behind (`non-spec:///x`) while ada::url leaves it untouched. Restore the saved_url snapshot the function already takes on entry, which is the same rollback every other setter in this file uses.

### DIFF
--- a/src/url.cpp
+++ b/src/url.cpp
@@ -696,9 +696,6 @@ bool url::set_host_or_hostname(const std::string_view input) {
 
   url saved_url(*this);
 
-  std::optional<std::string> previous_host = host;
-  std::optional<uint16_t> previous_port = port;
-
   size_t host_end_pos = input.find('#');
   std::string _host(input.data(), host_end_pos != std::string_view::npos
                                       ? host_end_pos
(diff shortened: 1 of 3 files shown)


anonrig (approved):

anonrig:
fuzzer fail with this change. can you take a look?

Arawoof06:
Good catch. The fuzzer wasn't unhappy about the aggregator, it was the aggregator now disagreeing with `ada::url`.

Turns out `url::set_host_or_hostname` had the same latent bug I fixed here: on a failed host parse it rolled back with `host = previous_host` + `update_base_port`, but `parse_host` sets `is_valid = false` and that partial rollback never restores it. So after a rejected `set_host`, `url::set_port` early-returns on the stale `is_valid == false` and silently leaves the port unchanged. The aggregator now snapshot-restores and stays valid, so the two diverged (e.g. failed `set_host("7\x03")` then `set_port("1")`: url kept 8080, aggregator went to 1).

Fixed url the same way: restore the `saved_url` snapshot it already captures on entry. Both backends now agree, the parse fuzzer runs clean past 1.2M execs, and all 295 ctest tests pass. Added a typed regression that a rejected set_host doesn't poison a later set_port.

lemire (approved):

PR #1170, opened 2026-07-07
reject overlong and surrogate utf-8 in tokenizer decoder

Arawoof06:
Tokenizer::get_next_code_point only rejects overlong two-byte leads and four-byte leads above 0xF4, so overlong three and four-byte encodings, UTF-16 surrogates, and code points past U+10FFFF still decode. That means E0 80 A8 turns into U+0028 and the tokenizer reads it as a literal '(' opening a regexp group, so a pattern with ill-formed bytes tokenizes differently from the same characters written plainly. Constrain the first continuation byte per RFC 3629 (E0<A0, ED>9F, F0<90, F4>8F) so those sequences are flagged invalid alongside the other malformed cases already handled here.

### DIFF
--- a/include/ada/url_pattern_helpers-inl.h
+++ b/include/ada/url_pattern_helpers-inl.h
@@ -409,6 +409,20 @@ constexpr void Tokenizer::get_next_code_point() {
     return;
   }
 
+  // Constrain the first continuation byte per RFC 3629 so overlong encodings,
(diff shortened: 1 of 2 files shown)


lemire:
The PR looks to be AI generated. You don’t show that you studied the public API and found a bug. Closing.

PR #1174, opened 2026-07-10
Expose retained memory estimates

mjackson:
- Add allocation-free retained-capacity estimates for `url_aggregator` and `url_search_params`
- Expose equivalent C APIs that include each opaque handle's owning result block
- Use saturating arithmetic so embedders can safely report external memory to their garbage collector

This lets embedders account for Ada-owned URL and query-list storage without inspecting private layout or allocating during GC.

### DIFF
--- a/include/ada/url_aggregator.h
+++ b/include/ada/url_aggregator.h
@@ -111,6 +111,17 @@ struct url_aggregator : url_base {
   [[nodiscard]] constexpr size_t get_href_size() const noexcept;
 
   /**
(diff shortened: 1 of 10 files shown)


anonrig (approved):
LGTM. Thank you for the PR

lemire:
@mjackson What is the purpose here? 

Will every little component of out software systems have a memory estimator bundled with it? 

What function does the estimator serve?

How often are systems memory limited because they are storing too much data related to URLs?

Are you going to be using this personally in a significant manner? Can you describe the usage you have in mind?

PR #1175, opened 2026-07-11
perf: speed up absolute http(s) parse for url and url_aggregator

anonrig:
## Summary
- Add `try_parse_simple_absolute`: a shared, table-driven single-pass fast path for absolute `http`/`https` URLs that are already normalized (clean ASCII host, no credentials/port/IPv4/IDNA, no path `%` / `.` / `..` segments). It runs before `has_tabs_or_newline` and fills either `url_aggregator` (resize + memcpy + component offsets) or `ada::url` (component strings).
- Optimize `ada::url::get_href` for the common special-scheme case with pre-sized buffer + `memcpy` (avoids temporary concatenations like `"?" + query`).
- Use a sized `string_view` needle for `xn-` host scans (avoid `strlen` on a C string).
- Add `simple_absolute_fast_path` regression tests covering empty-path slash insertion, percent-encoded query, `%2e` / `%2e%2e` fall-through, host lowercasing, credentials/port fall-through, and url vs aggregator href agreement.

## Performance
On Release `benchdata` (`BenchData_BasicBench_AdaURL_href` / `..._aggregator_href`), dual runs with `--benchmark_min_time=3s` vs a clean HEAD baseline:

| Type | Baseline mean | After mean | Improvement |
|------|---------------|------------|-------------|
| `ada::url` | ~136.0 ns/url | ~106.7 ns/url | **~21.6%** |
| `ada::url_aggregator` | ~89.5 ns/url | ~69.6 ns/url | **~22.2%** |

Interleaved A/B pairs also stayed ≥20% for both types.

## Test plan
- [x] `ctest` with `ADA_TESTING=ON` (full suite)
- [x] `basic_tests` filters: `*simple_absolute_fast_path*`, `*nodejs4*`
- [x] Release `benchdata` dual baseline vs after on Ada href filters
- [ ] CI green on this PR

### DIFF
--- a/fuzz/ada_c.cc
+++ b/fuzz/ada_c.cc
@@ -1,5 +1,10 @@
-// C API fuzzer. Amalgamates ada.cpp (same pattern as parse.cc).
-// Do not include ada_c.h — types are defined in the amalgamation.
+// C API fuzzer. Built as C++ amalgamating ada.cpp so OSS-Fuzz coverage
+// instrumentation sees both the harness and the library (the previous
+// separate-object C link left ada_c at ~0% coverage once hard aborts on
+// can_parse/parse mismatches stopped corpus progress).
+//
+// Types/functions from ada_c.cpp are already in this TU via the amalgamation;
+// do not also include ada_c.h (redefinition errors).
 
 #include <fuzzer/FuzzedDataProvider.h>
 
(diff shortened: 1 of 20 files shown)


lemire (approved):
We had a similar idea in the past...  so it is nice.

HOWEVER, I would make it 'opt in' and mark it as EXPERIMENTAL until it has been thoroughly fuzzed.

anonrig:
@lemire I'll add a lot more fuzzing to this PR so we can pass the experimental stuff.

howardpen9:
Nice work

PR #1179, opened 2026-07-13
perf: share and speed up IPv4/IPv6 host parsers

anonrig:
## Summary

Shared, faster IPv4/IPv6 host parsing used by both `ada::url` and `ada::url_aggregator`.

- **Shared `ada::ip_address` module** (`ip_address.h` / `ip_address-inl.h` / `ip_address.cpp`) so `url` and `url_aggregator` no longer duplicate large parse_ipv4/parse_ipv6 implementations (~600 lines removed from the two files).
- **IPv4 general path**: hand-rolled decimal / hex / octal segment parsers (no `std::from_chars`), with the existing pure-decimal fast path kept as the hot path plus a length early-out (7–16).
- **IPv6**: raw pointers, `int` compress sentinel instead of `std::optional`, hex nibble LUT, and one-pass compress expansion (no swap loop).
- **Serialization**: digit-pair LUT for IPv4 and nibble LUT for IPv6; write into stack buffers; aggregator updates hostname from `string_view` without temporary `std::string` growth where possible; skip rewrite when IPv6 hostname is already canonical.
- **Benchmarks**: `bench_ipv4` now includes IPv6 workloads (`Bench_IPv6_*`).

## Performance notes (Apple M-series, Release)

| Workload | `ada::url` (approx) |
|----------|---------------------|
| IPv4 decimal | ~39 ns/url |
| IPv4 non-decimal | ~75 ns/url |
| IPv6 mixed | ~50 ns/url |

Decimal stays in the same ballpark as the previous dedicated fast path. Main wins are shared code quality, non-decimal IPv4 without `from_chars`, IPv6 structure, and lower serialization cost in the aggregator.

## Test plan

- [x] `cmake -B build -G Ninja -DADA_TESTING=ON && cmake --build build`
- [x] `ctest --output-on-failure --test-dir build` (299/299)
- [x] `bench_ipv4` Release build with IPv4 decimal / non-decimal / IPv6 filters
- [ ] CI WPT + fuzzers


lemire:
@anonrig For AVX-512 capable systems, we can do much better, I have been working on this (I even have at least one blog post).

anonrig:
> @anonrig For AVX-512 capable systems, we can do much better, I have been working on this (I even have at least one blog post).

@lemire I've added

lemire (approved):

PR #1182, opened 2026-07-14
fix: avoid out-of-bounds read on empty input in is_ipv4

Arawoof06:
checkers::is_ipv4 calls view.back() before checking for an empty view, so an empty string_view reads one byte out of bounds (UBSAN reports an offset of SIZE_MAX applied to a null pointer). is_ipv4 is a public helper in checkers.h and fuzz/serializers.cc reaches it with fuzzer-controlled input. verify_dns_length in the same file already guards the empty case, so add the same check to is_ipv4.

### DIFF
--- a/fuzz/serializers.cc
+++ b/fuzz/serializers.cc
@@ -184,20 +184,23 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
     volatile bool has_tn = ada::unicode::has_tabs_or_newline(util_input);
     (void)has_tn;
 
-    // is_ipv4: must not crash on arbitrary input. Cross-check against URL
-    // parsing: if is_ipv4 reports true, embedding the string as a hostname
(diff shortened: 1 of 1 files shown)


anonrig (approved):

lemire:
@anonrig Here is the specification of the function :

```cpp
/**
 * @private
 * Returns true if an input is an ipv4 address. It is assumed that the string
 * does not contain uppercase ASCII characters (the input should have been
 * lowered cased before calling this function) and is not empty.
 */
```

Arawoof06:
Good point, the contract does say non-empty so is_ipv4 itself isn't wrong here. The only caller feeding it an empty view is the serializers fuzz harness (ConsumeRandomLengthString can return ""), which violates that precondition. I've dropped the guard from is_ipv4 and instead skip the empty case in the harness, so the documented precondition holds and the hot path stays untouched.

anonrig (approved):

PR #1185, opened 2026-07-15
bound hex ipv4 number by value not digit count in parse_ipv4_number

Arawoof06:
parse_ipv4_number caps a hex segment at eight digits, but leading zeros are allowed in a hex IPv4 number so a value that still fits in 32 bits can use more digits than that. Because of the cap, http://0x0000000ff/ (value 255) is now rejected even though the std::from_chars(radix 16) path it replaced in #1179 accepted it and returned host 0.0.0.255. This swaps the digit-count cap for the same value-range guard the octal branch already uses, so leading-zero hex numbers parse again while numbers above 0xFFFFFFFF stay rejected.


lemire (approved):
`parse_ipv4_number` is not in a public API.

lemire:
The issue is valid and the fix looks good.

PR #1186, opened 2026-07-16
fix no scheme state accepting any input containing a fragment

Arawoof06:
no scheme state decides whether a relative input may resolve against a base with an opaque path by testing `!fragment.has_value()`, but `prune_hash` strips the fragment from `url_data` before the state machine runs, so that only asks whether the input contains a '#' anywhere rather than whether c is '#'. Anything carrying a fragment therefore skips the failure branch and falls through to relative state, so `parse("//evil.com/p#", base="about:blank")` gives `about://evil.com/p#` with host `evil.com` and `parse("..#", base="a:b")` gives `a:b/#`, where Chrome, whatwg-url and the standard all return failure; the same inputs without the trailing '#' are already rejected, so one '#' is enough to grow an authority on an opaque-path base. Comparing `input_position` against `input_size` is what the fragment branch just below already does, and it is the only condition under which c can be '#' once the fragment has been pruned.

The `..#` expectation in `ada_extra_urltestdata.json` and in `node_issue_47889` came from #368, which fixed the abort in nodejs/node#47889 but settled on `a:b/#` where that issue asked for `ERR_INVALID_URL`; both now expect failure, and the test keeps its original point that the parser reports rather than crashes. Full suite is green and the fix moves every case I could find onto Chrome's answer.

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -526,10 +526,14 @@ result_type parse_url_impl(std::string_view user_input,
       }
       case state::NO_SCHEME: {
         ada_log("NO_SCHEME ", helpers::substring(url_data, input_position));
+        // The fragment was pruned from url_data before the state machine ran,
+        // so 'c is U+0023 (#)' holds exactly when a fragment was found and
+        // nothing else remains in front of it.
+        const bool c_is_hash =
+            fragment.has_value() && input_position == input_size;
         // If base is null, or base has an opaque path and c is not U+0023 (#),
         // validation error, return failure.
-        if (base_url == nullptr ||
-            (base_url->has_opaque_path && !fragment.has_value())) {
+        if (base_url == nullptr || (base_url->has_opaque_path && !c_is_hash)) {
           ada_log("NO_SCHEME validation error");
           url.is_valid = false;
           return url;
@@ -537,8 +541,7 @@ result_type parse_url_impl(std::string_view user_input,
         // Otherwise, if base has an opaque path and c is U+0023 (#),
         // set url's scheme to base's scheme, url's path to base's path, url's
         // query to base's query, and set state to fragment state.
-        else if (base_url->has_opaque_path && fragment.has_value() &&
-                 input_position == input_size) {
+        else if (base_url->has_opaque_path) {
           ada_log("NO_SCHEME opaque base with fragment");
           url.copy_scheme(*base_url);
           url.has_opaque_path = base_url->has_opaque_path;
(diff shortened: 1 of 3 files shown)


tests/wpt/ada_extra_urltestdata.json
```diff
@@ -274,16 +274,31 @@
...
+    "base": "a:b",
+    "href": "a:b#f",
     "origin": null,
     "protocol": "a:",
     "username": "",
     "password": "",
     "host": "",
     "hostname": "",
     "port": "",
-    "pathname": "b/",
```
anonrig, line 285: This seems like a breaking change.

src/parser.cpp
```diff
@@ -526,19 +526,22 @@ result_type parse_url_impl(std::string_view user_input,
...
           ada_log("NO_SCHEME validation error");
           url.is_valid = false;
           return url;
         }
         // Otherwise, if base has an opaque path and c is U+0023 (#),
         // set url's scheme to base's scheme, url's path to base's path, url's
         // query to base's query, and set state to fragment state.
-        else if (base_url->has_opaque_path && fragment.has_value() &&
-                 input_position == input_size) {
+        else if (base_url->has_opaque_path) {
```
anonrig: This path should also check for `and c is U+0023 (#)`

Arawoof06: Added in `1f69382`. You're right—I added the explicit `c == '#'` check. It was already implied by the previous failure branch, but spelling it out makes it match the spec more closely and makes the code a bit easier to follow.

tests/wpt/ada_extra_urltestdata.json
Arawoof06: Yeah, it is a behavior change.

The current expectation comes from #368. That fixed the abort in `nodejs/node#47889`, but the expected result (`a:b/#`) only works because of the parser bug.

With this change, the behavior matches Chrome, `whatwg-url`, and the URL Standard. In the `no scheme state`, if the base has an opaque path and the current character isn't `#`, parsing should fail. For `..#`, the current character is the first `.`, so it should return failure instead of `a:b/#`.

If you'd rather keep the parser fix and the test expectation update separate, I'm happy to split them into two PRs.

anonrig: I think this is the correct approach. Next release will be a major so we won't have issues with introducing breaking changes.

anonrig (approved):

PR #1187, opened 2026-07-17
reject unterminated group in url pattern constructor string parser

Arawoof06:
constructor_string_parser::parse increments the group depth on an open token and then falls through into the group-depth branch instead of continuing, so when "{" is the final token the end token gets consumed as group content and the loop walks off the token list without ever reaching the end handling that changes state to DONE. The component under construction is never assigned and quietly keeps its "*" default, so ada::parse_url_pattern("https://example.com{") returns a pattern whose hostname is "*" and matches any host, and "https://example.com/a{" returns pathname "*", where the string should instead fail to compile. Step 3 of parse a constructor string ends with "Continue."; this restores it, and the trailing component is then captured and compiled like any other.


anonrig (approved):
Nice catch

PR #1188, opened 2026-07-18
report capture groups for empty input in regex_search

Arawoof06:
std_regex_provider::regex_search returns an empty groups vector whenever the component input is the empty string, skipping group extraction before it starts. A URLPattern component whose regexp has a capture group that matches an empty value then loses that group from exec(): {search: "(x*)"} against an empty search reports {} instead of {"0": ""}, and the optional {search: ":a?"} reports {} instead of leaving the group undefined, while Chrome and the create-a-component-match-result algorithm keep the key. Dropping the input.empty() term runs the same 1..n extraction for an empty match; a successful regex_search always leaves the full match at index 0 so match_result is never empty here, and the FULL_WILDCARD fast path is untouched.

### DIFF
--- a/src/url_pattern_regex.cpp
+++ b/src/url_pattern_regex.cpp
@@ -38,8 +38,11 @@ std_regex_provider::regex_search(std::string_view input,
     return std::nullopt;
   }
   std::vector<std::optional<std::string>> matches;
-  // If input is empty, let's assume the result will be empty as well.
-  if (input.empty() || match_result.empty()) {
+  // An empty input can still match with capture groups (e.g. "(x*)" or an
+  // optional ":a?" against ""), so the group values must be extracted here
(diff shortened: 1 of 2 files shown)


anonrig (commented):
This might be true for std regex provider but might not be true for v8 based regex providers. right?

Arawoof06:
Right, this is std-provider-only. The concept just requires regex_search to hand back the capture group values, and a V8-backed provider builds that vector straight out of RegExpExec, so there's no equivalent early return to hit. Node already reports {"0": ""} for {search: "(x*)"} against an empty search, which is what I checked the expected output against.

create_component_match_result is untouched, so nothing shifts for other providers. This only brings std_regex_provider in line with what a V8-based one already does.

anonrig (approved):

PR #1193, opened 2026-07-27
fix: enforce max input length on href_from_file and search params

anonrig:
## Summary
- Make `href_from_file` respect `get_max_input_length()` for both the input path and the final `file://` href (returns empty string when rejected).
- Make `url_search_params` construction / `reset` refuse query strings longer than the same limit (object stays empty).
- Add regression tests covering overlength input, percent-encoding expansion, and the under-limit success path.

## Test plan
- [x] `./build/tests/max_input_length`
- [x] `./build/tests/from_file_tests`
- [x] `./build/tests/url_search_params`
- [ ] CI on the PR

Made with [Cursor](https://cursor.com)


lemire (commented):
ah!

PR #1195, opened 2026-07-28
perf: use SHRN-based zero test in the NEON kernels

lemire:
On arm64, testing whether a comparison mask is all-zero can be done by narrowing each 16-bit lane by four bits (SHRN) and comparing the resulting 64-bit value against 0.0 as a double. That is cheaper than a horizontal maximum (UMAXV) followed by a transfer to a general-purpose register, because it stays in the FP/SIMD unit on the no-match path. The narrowed value doubles as a position mask with four bits per input byte, so the same value that answers "is there a match?" also answers "where?".

has_tabs_or_newline already accumulates comparison results, so the new test drops straight in.

find_next_host_delimiter and find_next_host_delimiter_special did not have a comparison mask to work with: `classify` held 0x01/0x02/0x04 for a match, and SHRN keeps the high nibble of even-indexed bytes, which is zero for all of those values. Replacing the AND with vtstq_u8 (CMTST, same single instruction) yields a proper 0x00/0xFF mask, and the whole to_bitmask helper -- an AND plus three VPADDQ plus an extract -- along with the extra compare-against-zero then becomes unnecessary.

The floating-point compare is only valid for 0x00/0xFF lanes: each byte of the narrowed value is then 0x00, 0x0F, 0xF0 or 0xFF, so it can never form the negative-zero bit pattern, which would compare equal to 0.0.

Apple M-series, best of 7:
  find_next_host_delimiter_special, ~200-500 byte hosts: 1.3-1.5x faster
  find_next_host_delimiter_special, typical short hosts: ~1.04x (noise)
  has_tabs_or_newline: neutral -- the test is outside the loop

### DIFF
--- a/src/helpers.cpp
+++ b/src/helpers.cpp
@@ -180,6 +180,47 @@ ada_really_inline int trailing_zeroes(uint32_t input_num) noexcept {
 #endif  // ADA_REGULAR_VISUAL_STUDIO
 }
 
+#if ADA_NEON
+// computes the number of trailing zeroes of a 64-bit word
+// this is a private inline function only defined in this source file.
+ada_really_inline int trailing_zeroes64(uint64_t input_num) noexcept {
+#ifdef ADA_REGULAR_VISUAL_STUDIO
+  unsigned long ret;
+  _BitScanForward64(&ret, input_num);
+  return (int)ret;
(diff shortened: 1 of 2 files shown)


anonrig (approved):
nice work!

PR #1197, opened 2026-07-31
Speed up can_parse for canonical HTTP(S) URLs

ImLunaHey:
## Summary

This ports the canonical HTTP(S) validation fast path from [Lucid URL](https://github.com/lucid-softworks/url), our URL library, into Ada's `can_parse` implementation.

It adds a narrow `can_parse` front end for already-canonical lowercase HTTP(S) URLs. The new path:

- recognizes exact lowercase `http://` and `https://` inputs
- scans clean host bytes eight at a time using a lookup table
- falls back to the existing validator for ports, credentials, IPv4, IDNA, uppercase hosts, normalization, and other complex cases
- preserves the existing normalized-output length checks

## Why

The existing validation fast path still performs broader scheme and authority processing for the overwhelmingly common canonical HTTP(S) case. Handling that conservative subset first avoids unnecessary work while retaining the current behavior for everything outside the subset.

On the shared 100,025-URL benchmark corpus used to compare Ada with the source implementation in Lucid URL, median `can_parse` performance changed from:

| | Median time | Throughput |
|---|---:|---:|
| Ada baseline | 36.77 ns/URL | 27.2M URLs/s |
| This PR | 12.75 ns/URL | 78.4M URLs/s |
| Lucid URL | 11.09 ns/URL | 90.2M URLs/s |

That is a 2.88x speedup for Ada, or a 65.3% reduction in median time per URL. Parsing and serialization paths are unchanged.

## Validation

- All 318 tests pass.
- Added focused `can_parse`/full-parser consistency coverage for fast-path and fallback inputs.
- The shared benchmark corpus reports zero validity or serialization disagreements.
- `bash tools/run-clangcldocker.sh` passes (clang-format and clang-tidy).

### DIFF
--- a/src/implementation.cpp
+++ b/src/implementation.cpp
@@ -1,5 +1,6 @@
 #include "ada/implementation-inl.h"
 
+#include <array>
 #include <atomic>
 #include <limits>
 #include <optional>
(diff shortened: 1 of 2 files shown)


anonrig (approved):

anonrig:
Thank you

PR #1199, opened 2026-07-31
perf: widen simple-absolute http(s) parse fast path

anonrig:
## Summary
Widen the existing `try_parse_simple_absolute` hot path for absolute `http`/`https` (no base URL):

- Bulk host / rest validation (NEON on Apple Silicon where available)
- `memchr` locate for `?` / `#`
- Fail-closed host checks (`is_ipv4`, punycode) and path `.` / `..` / `%2e` rejection
- Cheap digit / `[` peek so pure IPv4/IPv6 inputs never enter the fast path

On success, fill `url` components and the `url_aggregator` buffer directly (no freelist, no href cache).

## Scope
- `src/parser.cpp`, `tests/basic_tests.cpp`, small comment in `src/implementation.cpp`
- **No** `string_pool`, **no** `url` href cache, **no** public API signature changes

## Independence
Targets **`main` only**. Independent of the get-href and href-cache PRs (not stacked).

## Test plan
- [x] `basic_tests` simple-absolute cases
- [ ] Full CI
- [ ] Optional: Release `benchdata` filter for aggregator/url href

PR #1200, opened 2026-07-31
perf: speed up url::get_href for common special URLs

anonrig:
## Summary
Speed up `ada::url::get_href()` when the URL is special, has a host, and has no credentials or port:

- Prefer uninitialized string resize (`resize_and_overwrite` / Apple `__resize_default_init` / fallback `resize`)
- Hard-code `https` / `http` scheme prefixes to avoid special-scheme table loads

## Scope
- `include/ada/url-inl.h` only
- **No** parser changes

## Independence
Targets **`main` only**. Independent of the simple-absolute and href-cache PRs; can land in any order (not stacked).

## Test plan
- [ ] Full CI
- [ ] Optional: local Release microbench on `AdaURL_href`

### DIFF
--- a/include/ada/url-inl.h
+++ b/include/ada/url-inl.h
@@ -185,40 +185,75 @@ constexpr void url::copy_scheme(const ada::url& u) {
   type = u.type;
 }
 
+namespace detail {
+// Grow string to n bytes without requiring value-init of new chars when the
+// platform provides that API. Not noexcept: allocation may throw bad_alloc.
+ada_really_inline void string_resize_uninitialized(std::string& s, size_t n) {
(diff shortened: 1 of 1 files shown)


anonrig:
Closing: CodSpeed reports a **−3.06%** regression on `BenchData_BasicBench_AdaURL_href` (98.1 → 101.2 ms) vs main.

On Linux/libstdc++ the uninitialized-resize path often falls back to `resize` (same zero-init as before) while adding scheme-branch overhead, so this change is net-negative in Simulation and is not worth landing.

The larger wins for that bench already land via:
- #1199 (simple-absolute parse, ~+15% on the same CodSpeed metric)
- #1201 (href cache + pool when needed)

No further work planned on this branch.

PR #1124, opened 2026-04-06
optimize percent_encode, percent_encode_index, and percent_decode

thisalihassan:
- add SIMD fast paths for `percent_encode_index`, `percent_encode`, and `percent_decode`
- add direct unicode tests for boundary lengths, malformed `%` sequences, non-ASCII input, and setter paths
- extend `benchmarks/percent_encode.cpp` with longer inputs and decode coverage

### Release build on arm64:

- `Fragment`: `91.443 ns -> 72.306 ns` (`1.265x`)
- `Query`: `90.308 ns -> 73.698 ns` (`1.225x`)
- `SpecialQuery`: `90.111 ns -> 72.523 ns` (`1.243x`)
- `UserInfo`: `256.253 ns -> 192.942 ns` (`1.328x`)
- `C0Control`: `88.458 ns -> 73.321 ns` (`1.206x`)

Tests written by Codex & Claude

Closes [#1120](https://github.com/ada-url/ada/issues/1120).

PR #1202, opened 2026-07-31
perf: speed up percent_decode using bulk run-copying

jbergstroem:
Speed up `ada::unicode::percent_decode`, the routine that decodes `%XX` escapes when parsing hosts (via `to_ascii`) and search-param values.

The decoded output is never longer than the input, size the result buffer once and copy plain runs in bulk; the same approach `form_urlencoded_decode` already uses.

## What changed

- Allocate the output once at `input.size()` and shrink to the final length at the end, instead of `reserve` +  per-byte `+=`.
- Plain (non-`%`) spans are located with `std::memchr('%')` and copied with a single `std::memcpy`, rather than appending one character at a time.
- `%XX` runs are decoded through the existing `unhex_table` (two loads + an OR to validate), which is above the function so both `percent_decode` and `form_urlencoded_decode` share it.
- Add `benchmarks/percent_decode.cpp`, mirroring `percent_encode.cpp`.
- Increase coverage in the test suite to pick up introduced branches.

## Benchmark results

Measured with a Apple M5 (`benchmarks/percent_decode`, `--benchmark_min_time=1s`):

| Benchmark | Before | After | Improvement |
|-----------|--------|-------|-------------|
| `Decode` time/url | ~54.3 ns | ~36.4 ns | **1.49×** |
| `Decode` throughput | ~1.03 GB/s | ~1.54 GB/s | **1.49×** |

## Testing

- Full test suite passes
- `clang-format`, `clang-tidy` introduces no new findings
- Verified output to be byte-identical to the previous implementation over fuzzed inputs plus a few hand-picked edge cases (`%`, `%2`, `%%41`,`%g0`, trailing `%`, doubly-encoded, some utf-8 strings)

Edit: side-note, just noticed https://github.com/ada-url/ada/pull/1124 which might be a longer term solution; feel free to close this if thats the preferred path.

### DIFF
--- a/src/unicode.cpp
+++ b/src/unicode.cpp
@@ -455,22 +455,6 @@ unsigned constexpr convert_hex_to_binary(const char c) noexcept {
   return hex_to_binary_table[c - '0'];
 }
 
-// 0..15 for hex digits, 0xFF otherwise - validate and decode with two loads.
-constexpr static std::array<uint8_t, 256> unhex_table = []() consteval {
-  std::array<uint8_t, 256> t{};
-  for (size_t i = 0; i < 256; ++i) {
-    t[i] = 0xFF;
-  }
-  for (uint8_t i = 0; i < 10; ++i) {
-    t[static_cast<size_t>('0') + i] = i;
-  }
-  for (uint8_t i = 0; i < 6; ++i) {
-    t[static_cast<size_t>('A') + i] = static_cast<uint8_t>(10 + i);
-    t[static_cast<size_t>('a') + i] = static_cast<uint8_t>(10 + i);
-  }
-  return t;
-}();
-
 std::string percent_decode(const std::string_view input, size_t first_percent) {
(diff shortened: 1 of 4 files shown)


anonrig:
Nice work!

src/unicode.cpp
```diff
@@ -504,6 +471,60 @@ constexpr static std::array<uint8_t, 256> unhex_table = []() consteval {
   return t;
 }();
 
+std::string percent_decode(const std::string_view input, size_t first_percent) {
```
anonrig, line 458: Can you put the method to the previous place so the git diff is nicer?

anonrig:
>Edit: side-note, just noticed https://github.com/ada-url/ada/pull/1124 which might be a longer term solution; feel free to close this if thats the preferred path.

It has regressions... That's why we couldn't land it.

anonrig (approved):

PR #1203, opened 2026-07-31
perf: skip setter rollback snapshot when length limit is unreachable

jbergstroem:
Another quick win found while hiking through the codebase:

Most `url_aggregator` setters copy the entire buffer up front (`url_aggregator saved_url(*this)`) purely so they can restore it if the mutation pushes the URL past `get_max_input_length()`. That limit defaults to ~4 GB, so in practice the rollback never fires and the copy is dead weight on every `set_*` call.

These are obviously not in the hot path, but a win none the less.

## What changed

Introduce a new guard:

```cpp
[[nodiscard]] bool needs_rollback_snapshot(size_t input_len) const noexcept {
  return buffer.size() + input_len * 3 + 16 > ada::get_max_input_length();
}
```

A setter grows the buffer by at most `input_len * 3` (worst-case percent-encoding) plus a few bytes of delimiters, so when that upper bound still fits under the limit the length-guard rollback cannot fire and the snapshot is skipped. When it is needed, it's now taken lazily via `std::optional<url_aggregator>`.

Applied to the six setters whose rollback is limited by length: `set_protocol`, `set_username`, `set_password`, `set_pathname`, `set_search`, `set_hash`.

Intentionally skipped: `set_port` and `set_host_or_hostname`. Their rollbacks also handle `parse_port` / `parse_host` failures, which are not length-related, so the snapshot there is still required.

Finally, add `benchmarks/bench_setters.cpp`, which benchmarks every setter to make the case. I can fold these someplace else given their importance?

## Benchmarks

Measured with `bench_setters` with a Apple M5:

| Setter | Before | After | Speedup |
|--------|--------|-------|---------|
| `set_username` | 22.1 ns | 11.0 ns | **2.01×** |
| `set_protocol` | 29.1 ns | 18.3 ns | **1.59×** |
| `set_password` | 30.0 ns | 19.5 ns | **1.54×** |
| `set_hash` | 26.9 ns | 18.9 ns | **1.42×** |
| `set_search` | 36.6 ns | 26.3 ns | **1.39×** |
| `set_pathname` | 49.6 ns | 39.8 ns | **1.25×** |

`set_port`, `set_host_or_hostname`, and `set_href` are unchanged.

## Testing

- Full suite passes
- `clang-format`, `clang-tidy` are happy

### DIFF
--- a/include/ada/url_aggregator.h
+++ b/include/ada/url_aggregator.h
@@ -390,6 +390,17 @@ struct url_aggregator : url_base {
    */
   [[nodiscard]] constexpr bool cannot_have_credentials_or_port() const;
 
+  /**
+   * @private
+   * Several setters restore a saved copy of the URL when the mutation pushes
+   * the buffer past get_max_input_length(). A setter grows the buffer by at
+   * most input_len * 3 (worst-case percent-encoding) plus a small constant for
+   * delimiters, so this returns false when that upper bound stays within the
+   * limit. It does not account for parse-failure rollbacks, which are
+   * unrelated to the length limit.
+   */
+  [[nodiscard]] bool needs_rollback_snapshot(size_t input_len) const noexcept;
+
   template <bool override_hostname = false>
   bool set_host_or_hostname(std::string_view input);
 
--- a/src/url_aggregator.cpp
+++ b/src/url_aggregator.cpp
@@ -210,6 +210,13 @@ inline void url_aggregator::set_scheme(std::string_view new_scheme) {
   ADA_ASSERT_TRUE(validate());
 }
 
+bool url_aggregator::needs_rollback_snapshot(size_t input_len) const noexcept {
+  // 3x should cover worst-case percent-encoding of every input byte; the
+  // constant covers the handful of delimiter bytes a setter may add ("/.", "?",
+  // "#" and so on)
+  return buffer.size() + input_len * 3 + 16 > ada::get_max_input_length();
+}
+
 bool url_aggregator::set_protocol(const std::string_view input) {
   ada_log("url_aggregator::set_protocol ", input);
   ADA_ASSERT_TRUE(validate());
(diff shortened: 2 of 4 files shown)


include/ada/url_aggregator.h
```diff
@@ -390,6 +390,16 @@ struct url_aggregator : url_base {
    */
   [[nodiscard]] constexpr bool cannot_have_credentials_or_port() const;
 
+  /**
+   * Several setters restore a saved copy of the URL when the mutation pushes
```
anonrig, line 395: This should have `@private`

anonrig:
can you fix the conflicts as well?

jbergstroem:
windows fail seems unrelated.

src/url_aggregator.cpp
```diff
@@ -210,6 +210,13 @@ inline void url_aggregator::set_scheme(std::string_view new_scheme) {
   ADA_ASSERT_TRUE(validate());
 }
 
+bool url_aggregator::needs_rollback_snapshot(size_t input_len) const noexcept {
+  // 3x should cover worst-case percent-encoding of every input byte; the
+  // constant covers the handful of delimiter bytes a setter may add ("/.", "?",
+  // "#" and so on)
+  return buffer.size() + input_len * 3 + 16 > ada::get_max_input_length();
```
anonrig, line 217: Why did you add 16? Can you add a comment

jbergstroem: I explained it in greater length in the header definition: https://github.com/ada-url/ada/pull/1203/changes#diff-403afdb21d95e7eb802306a12c8aabc23d25efe9366fa168e5b64e09bfba1956R397-R399

want me to copy over?

anonrig:
@lemire would you mind reviewing?

anonrig (approved):

PR #1212, opened 2026-08-11
validate ipv4 group structure in avx-512 fast path

Arawoof06:
try_parse_ipv4_avx512 only checks that every host byte is a digit or dot and that the dot count is three, then hands the buffer to parse_ipv4_decimal_trusted, which assumes each group is one to three digits and non-empty. it isn't: a group longer than three digits like the host in http://1234.5.6.7/ has its fourth digit skipped as if it were the separator, and an empty group like http://1..2.34/ decodes the '.' as a digit, so both parse to a bogus in-range value and are accepted as ipv4 on avx-512 builds while the scalar parse_ipv4_decimal_scalar and the whatwg host parser reject them. the trusted convert also reads the first byte of each group with *p - '0' without a p<pend guard, unlike the scalar path. route the fast path through parse_ipv4_decimal_scalar on the stack copy so the avx-512 build accepts exactly the hosts the portable build does, and drop the now-unused trusted variant.

### DIFF
--- a/include/ada/checkers-inl.h
+++ b/include/ada/checkers-inl.h
@@ -94,35 +94,6 @@ parse_ipv4_decimal_scalar(const char* p, const char* pend) noexcept {
 }
 
 #if defined(ADA_AVX512)
-// After SIMD validation: fewer rejection branches on convert.
-ada_really_inline uint64_t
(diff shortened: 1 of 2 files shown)


lemire:
I like the new tests.

lemire (approved):

PR #1217, opened 2026-08-17
perf: take canonical IPv4 on the aggregator fast path

anonrig:
Independent of #1216 (nibble-table SIMD). This PR only lets `url_aggregator` finish already-canonical dotted-quad IPv4 on the prefix path.

**Why this should not regress the IPv4 benches that #1216's combined patch lost:**

- **`ada::url` is unchanged.** Digit-led hosts still skip the prefix path in both `parse()` and `try_parse_simple_absolute()`, same as main. Decimal AdaURL stays on the general parser (finishing IPv4 as `ada::url` was the -24% regression).
- **Non-decimal IPv4 stays off the prefix path.** Hex (`0x...`), octal, short form, and trailing dots fail `host_is_canonical_ipv4_shape` in a few bytes. `parse()` then skips `try_parse_simple_absolute` entirely, so NonDecimal Aggregator does not pay a host scan.
- **Decimal Aggregator** is the intended win: a 7-15 byte dotted quad is copied with the existing memcpy finish and `host_type = IPV4`.
- **No `-mssse3`.** Helpers and unicode stay SSE2.
- IPv4 plus a non-simple tail (embedded NUL, encoding) falls through so `ada::url` and `url_aggregator` keep the same href.

### DIFF
--- a/include/ada/parser.h
+++ b/include/ada/parser.h
@@ -83,7 +83,7 @@ bool finish_simple_absolute_with_port(std::string_view input, result_type& out,
                                       ada::scheme::type scheme_type,
                                       uint32_t protocol_end, size_t host_start,
                                       size_t host_end, size_t host_len,
-                                      bool has_upper);
+                                      bool has_upper, bool is_ipv4);
 
 /** @private */
 template <class result_type>
--- a/include/ada/url.h
+++ b/include/ada/url.h
@@ -388,7 +388,7 @@ struct url : url_base {
   template <class result_type>
   friend bool ada::parser::finish_simple_absolute_with_port(
       std::string_view, result_type&, ada::scheme::type, uint32_t, size_t,
-      size_t, size_t, bool);
+      size_t, size_t, bool, bool);
   template <class result_type>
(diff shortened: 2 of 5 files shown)


anonrig:

- `Bench_IPv4_Decimal_Aggregator` **-8.75%** (the intended win)
- `Bench_IPv4_NonDecimal_Aggregator` **-5.38%**
- `Bench_BasicBench_AdaURL_aggregator_href` **-4.31%**
- `SetHash` / `SetPassword` **-5.93% / -3.31%**

`parse_host` already has `try_parse_ipv4_fast`. Sending canonical quads through the prefix path (scheme match, host scan, last-label, shape check, then memcpy + inserted `/`) is more work than that kernel. The extra `parse()` skip-loop state also taxes every aggregator URL, which is the href/setter noise.

The earlier combined-PR Decimal Aggregator gain did not reproduce once nibble SIMD and `-mssse3` were removed. Closing this; IPv4 stays on the general parser. #1216 is the change that actually improves href/DNS.

PR #1218, opened 2026-08-17
perf: nibble-table SIMD scan for percent_encode_index

anonrig:
First follow-up to #1216, landed independently so CodSpeed can attribute this site on its own.

Port the oven-sh/WebKit#454 nibble-table classifier (`low[b & 0xF] & high[b >> 4]`) onto `percent_encode_index` for 16+ byte inputs. Short inputs stay on the existing inlined 8-byte unroll (setters).

- `consteval` tables from each 32-byte percent-encode bitset (`static_assert(tables.fits)`)
- SSSE3 `pshufb` / NEON `tbl` when available; remainder is a scalar walk from the last full 16-byte chunk
- `percent_encode` now starts from `percent_encode_index` instead of `std::ranges::find_if`
- SSSE3 is `target("ssse3")` on the scanner only (same enablement as #1216). clang-cl/MSVC stay on the scalar path because they ignore GNU `target("ssse3")`
- No IPv4 changes and no project-wide `-mssse3`

Forbidden-domain classification is a separate PR.

### DIFF
--- a/include/ada/unicode-inl.h
+++ b/include/ada/unicode-inl.h
@@ -6,6 +6,9 @@
 #define ADA_UNICODE_INL_H
 #include "ada/unicode.h"
 #include "ada/character_sets.h"
+#include "ada/character_sets-inl.h"
+
+#include <cstring>
 
 /**
  * Unicode operations. These functions are not part of our public API and may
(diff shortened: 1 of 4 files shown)


anonrig:

- `SetHash` **-6.79%**
- `UserInfo` **-6.06%**
- `SetPassword` **-4.87%**
- `SetHref` **-4.69%**

Setters and `UserInfo` are short strings. The existing inlined 8-byte unroll already wins there; the 16-byte dispatch plus an out-of-line nibble scan (table select, `pshufb`, scalar tail) is extra work on the common path. Closing this. #1216 is the nibble change that actually improves href/DNS.

PR #1219, opened 2026-08-17
perf: nibble-table SIMD scan for forbidden domain code points

anonrig:
Second follow-up to #1216, independent of #1218 so CodSpeed can attribute this site on its own.

Port the oven-sh/WebKit#454 nibble-table classifier (`low[b & 0xF] & high[b >> 4]`) onto `contains_forbidden_domain_code_point` / `_or_upper` for 16+ byte hosts. Inputs under 16 bytes stay on the existing 4-byte table unroll (`parse_host` of typical short names is unchanged).

- `consteval` tables from the existing 256-byte class array (`static_assert(tables.fits)`)
- ASCII uppercase is a range compare, not a second nibble table
- SSSE3 `pshufb` / NEON `tbl` when available; overlapping last 16 bytes after a full chunk
- SSSE3 is `target("ssse3")` on the scanner only (same enablement as #1216 / #1218). clang-cl/MSVC stay on the scalar path
- No IPv4 changes and no project-wide `-mssse3`

Most hosts are shorter than 16 bytes, so this is the smaller of the two follow-ups.

### DIFF
--- a/include/ada/unicode.h
+++ b/include/ada/unicode.h
@@ -88,8 +88,8 @@ ada_really_inline constexpr bool is_forbidden_host_code_point(char c) noexcept;
  * Checks if the input contains a forbidden domain code point.
  * @see https://url.spec.whatwg.org/#forbidden-domain-code-point
  */
-ada_really_inline constexpr bool contains_forbidden_domain_code_point(
-    const char* input, size_t length) noexcept;
+bool contains_forbidden_domain_code_point(const char* input,
+                                          size_t length) noexcept;
 
 /**
  * @private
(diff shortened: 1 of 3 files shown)


anonrig:

- `SetHash` **-4.51%**
- `SetPassword` **-3.31%**

Those benches do not parse 16+ byte hosts, so the nibble path never runs. The hit is code size in the unity TU (`unicode.cpp` is compiled into `ada.cpp`), which taxes unrelated setters. Typical hosts are already under 16 bytes and stay on the 4-byte table unroll. Closing this. #1216 is the nibble change that actually improves href/DNS.

PR #1222, opened 2026-08-17
perf: cut more parse instructions with 8-byte scheme match and SWAR

anonrig:
The previous fast-path work (#1221) is on `main`. This follow-up reduces instruction count further on already-canonical special-scheme URLs without changing the public API.

## Changes

**8-byte scheme match.** `https://` is compared as a single little-endian `uint64_t` load/`cmp`. Shorter special schemes (`http://`, `wss://`, `ftp://`, `ws://`) use the same register with a mask, so the common case does not walk scheme bytes.

**32-byte host scan.** Host classification now unrolls like path/query/hash. That helps BBC-style CDN hosts that miss a single 16-byte window. Uppercase and `'x'` flags only count host bytes, not path bytes that share the delimiter window.

**Cheaper host post-checks.** Punycode (`xn-`) is skipped unless the host scan saw an `x` or uppercase. Last-label IPv4 gating uses the existing inline `checkers::last_label_may_be_a_number`.

**`can_parse` SWAR.** The 8-byte clean-host check is integer ALU (range/`eq` SWAR) instead of eight table lookups, and the HTTP(S) prefix uses the same 8-byte compare. Equality is borrow-safe so `./` and `_^` are not treated as a match.

**Userinfo skip.** Credential URLs (including the SetHref bench) miss the absolute fast path. A 16-byte `'@'` vs `/?#` check skips the `never_inline` call so those URLs do not pay for a guaranteed miss. `'@'` after a path delimiter still takes the fast path.

## Tests

All 330 tests passed locally, including new cases for 32-byte hosts, `ws://`, `xn--` in the path, uppercase path on a short host, `can_parse` hosts with `_` / `~` / `-`, and `set_href` of the credential bench URL.

## Compatibility

No public method signatures change. Fast-path misses still fall through to the WHATWG state machine.

<div><a href="https://cursor.com/agents/bc-01a01151-4f85-7d70-9b4d-513acb2454e4?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-01a01151-4f85-7d70-9b4d-513acb2454e4&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/src/implementation.cpp
+++ b/src/implementation.cpp
@@ -42,12 +42,37 @@ constexpr std::array<uint8_t, 256> clean_http_host_byte = []() consteval {
   return result;
 }();
 
+// SWAR: eight ASCII host bytes in [a-z0-9-._~] without 8 table lookups.
+// High bits must be clear first so the range subtracts cannot borrow
+// across bytes.
+ada_really_inline uint64_t swar_byte_eq(uint64_t w, uint8_t c) noexcept {
+  constexpr uint64_t k_ones = 0x0101010101010101ull;
+  constexpr uint64_t k_high = 0x8080808080808080ull;
+  const uint64_t x = w ^ (k_ones * c);
+  return (x - k_ones) & ~x & k_high;
+}
+
+ada_really_inline uint64_t swar_in_range(uint64_t w, uint8_t lo,
+                                         uint8_t hi) noexcept {
+  constexpr uint64_t k_ones = 0x0101010101010101ull;
+  constexpr uint64_t k_high = 0x8080808080808080ull;
+  const uint64_t ge_lo = (w | k_high) - (k_ones * lo);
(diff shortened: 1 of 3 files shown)


lemire (approved):
Looks sensible at a glance.

PR #1226, opened 2026-08-19
fix is_already_nfc for a composite starter before a lower-class mark

Arawoof06:
ada::idna::normalize skips the decompose/compose passes whenever is_already_nfc reports the input is already NFC, and to_ascii/to_unicode lean on the same check. That quick check tracks canonical ordering using each character's own combining class, but a precomposed starter such as U+00E1 (a + U+0301, class 230) has class 0 while its canonical decomposition ends in a class-230 mark. When such a starter is followed by a combining mark of lower class, say U+0323 dot below at class 220, the string is not in NFC: full normalization decomposes the starter, reorders the dot ahead of the acute, and recomposes a+dot to U+1EA1 with the acute left floating. is_already_nfc reported NFC anyway, so the host passed through untouched and IDNA produced the wrong A-label, turning http://%C3%A1%CC%A3/ into xn--1ca07i instead of xn--lsa752l. I ran into it while diffing host serialization against another URL implementation over random Unicode labels. The fix adds a small helper returning the trailing combining class of a character's canonical decomposition and carries that in prev_ccc, so a composite starter ahead of a lower-class mark is caught; it can only defer to the existing full normalization, never skip it wrongly. I added a regression test over both backends and the full suite including wpt stays green.

### DIFF
--- a/src/ada_idna.cpp
+++ b/src/ada_idna.cpp
@@ -5330,6 +5330,22 @@ static bool would_compose(std::u32string_view input) noexcept {
   return false;
 }
 
+// Trailing canonical combining class of a character: the combining class of the
+// last code point of its canonical decomposition, or the character's own class
+// when it has none. A precomposed starter such as U+00E1 (a + U+0301, class
+// 230) has class 0 but a nonzero trailing class, so a following mark of lower
+// class reorders past the hidden mark during normalization.
+static uint8_t trailing_ccc(char32_t c) noexcept {
+  const size_t length = canonical_decomp_length(c);
+  if (length == 0) {
+    return get_ccc(c);
+  }
+  const uint16_t* const decomposition =
+      decomposition_block_row(decomposition_index[c >> 8]) + (c % 256);
+  const size_t base = decomposition[0] >> 2;
+  return get_ccc(decomposition_data[base + length - 1]);
+}
+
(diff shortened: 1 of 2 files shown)


src/ada_idna.cpp
```diff
@@ -5330,6 +5330,22 @@ static bool would_compose(std::u32string_view input) noexcept {
   return false;
 }
 
+// Trailing canonical combining class of a character: the combining class of the
+// last code point of its canonical decomposition, or the character's own class
+// when it has none. A precomposed starter such as U+00E1 (a + U+0301, class
+// 230) has class 0 but a nonzero trailing class, so a following mark of lower
+// class reorders past the hidden mark during normalization.
+static uint8_t trailing_ccc(char32_t c) noexcept {
```
anonrig, line 5338: the equivalent should open to ada-url/idna repository

Arawoof06: Good point, ada_idna.cpp is generated so this would get overwritten on the next import. Opened the same fix upstream in ada-url/idna#87 (with a to_ascii regression test there). Happy to close this one once that lands and gets re-vendored, or leave it as the tracking change here, whichever you prefer.

anonrig (approved):

PR #1229, opened 2026-08-20
perf: replace aggregator queries with one structural edit

dmmulroy:
## Summary

Replace or insert `url_aggregator` queries with one structural edit when a fragment follows the query.

Previously query replacement could erase the old query, insert `?`, and then insert the plain or encoded value, moving the fragment tail multiple times. This change:

- encodes into a temporary only when needed;
- keeps an existing `?` delimiter and replaces the query value with one `replace_and_resize` call; and
- inserts an absent `?` + query span with one structural insertion.

The benchmark now covers alternating query lengths, encoded queries, and insert/clear operations.

## Performance

Apple M4 Pro, Apple clang 21, Release build, 20 alternating A/B rounds. `url_aggregator::set_search` had the exact same address in both binaries for measurement only; alignment is not part of this PR.

| Workload | Paired median CPU-time delta |
|---|---:|
| Search, repeated equal length | **-27.64%** |
| Search, alternating lengths | **-10.76%** |
| Search, alternating encoded lengths | **-17.58%** |
| Search insert/clear | **-13.18%** |

## Correctness

Typed tests cover both URL representations and verify shorter, equal-length, longer, encoded, absent, and cleared queries while preserving credentials, path, and fragment tails.

- `ctest --output-on-failure --test-dir build`: **338/338 passed**
- `bash tools/run-clangcldocker.sh`: LLVM 22 format and tidy passed

There is no public signature or object-layout change.

### DIFF
--- a/benchmarks/bench_setters.cpp
+++ b/benchmarks/bench_setters.cpp
@@ -119,6 +119,36 @@ static void SetSearch(benchmark::State& state) {
 }
 BENCHMARK(SetSearch);
 
+static void SetSearchAlternatingLength(benchmark::State& state) {
+  auto url = ada::parse<ada::url_aggregator>(base_url).value();
+  bool use_long_value = false;
+  run_setter(state, [&] {
+    use_long_value = !use_long_value;
+    url.set_search(use_long_value ? "?a=much-longer-query-value" : "?x=1");
+  });
+}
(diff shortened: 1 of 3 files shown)


anonrig (approved):

anonrig:
@dmmulroy can you fix the conflicts?

anonrig (approved):

PR #1232, opened 2026-08-22
perf: parse dotted-decimal IPv4 with AVX-512VL notab5

lemire:
## Summary

The AVX-512BW+VL IPv4 fast path only checked that the host was digits and dots, then converted with the scalar parser. This replaces that with the table-free `parse_ipv4_avx512vl_notab5` kernel:

- Masked 16-byte load (no source over-read).
- Digit placement from compressed delimiter positions (`vpcompressb`).
- Octet overflow as a dword compare on the reversed zero-padded group, in parallel with the convert.
- Octets packed with `vpmovdb` + `bswap` (no stack `oct[4]`).
- `vpdpbusd` when `__AVX512VNNI__` is available, otherwise `pmaddubsw`/`pmaddwd`.
- One WHATWG trailing dot is stripped before the strict four-group parse.
- Octal, hex, leading zeros, and short forms still return `ipv4_fast_fail` so the general parser handles them.

The IPv4 kernel is gated on `ADA_AVX512 && __AVX512VBMI2__` (`vpcompressb` is not BW+VL). `ADA_AVX512` itself stays `__AVX512BW__ && __AVX512VL__` so IPv6 is unchanged. `-march=skylake-avx512` / `-mavx512bw -mavx512vl` compile again (scalar IPv4 fallback). Portable scalar path is unchanged.

## Performance

Xeon Gold 6548N, gcc 14.3.1, `-march=native` Release, `bench_ipv4` decimal URLs (re-run after VBMI2 gate + `vpmovdb`/`bswap`):

| Bench | Before (cycles/url) | After | ns/url | url/s |
|---|---|---|---|---|
| AdaURL | 317 | **311** | 90.8 → **89.0** | 10.5M → **11.1M** |
| Aggregator | 400 | **377** | 114.4 → **108.3** | 8.5M → **9.1M** |

Instructions/url on AdaURL: 1291 → 1162. Non-decimal IPv4 is unchanged within noise (those hosts miss the strict fast path on purpose).

## Tests

- Direct `try_parse_ipv4_fast` checks for packed values (`192.168.1.1` → `0xC0A80101`), trailing dot, overflow, leading zeros, and short form.
- Existing malformed-group tests still apply.
- On the same AVX-512 box: 203/203 `basic_tests`, WPT `urltestdata`.

### DIFF
--- a/include/ada/checkers-inl.h
+++ b/include/ada/checkers-inl.h
@@ -11,7 +11,7 @@
 #include "ada/checkers.h"
 #include "ada/common_defs.h"
 
-#if defined(ADA_AVX512)
+#if defined(ADA_AVX512) && defined(__AVX512VBMI2__)
 #include <immintrin.h>
 #endif
 
@@ -123,13 +123,14 @@ parse_ipv4_decimal_scalar(const char* p, const char* pend) noexcept {
   return ipv4;
 }
 
-#if defined(ADA_AVX512)
+#if defined(ADA_AVX512) && defined(__AVX512VBMI2__)
 // Table-free AVX-512VL IPv4 parse (simdip parse_ipv4_avx512vl_notab5).
-// Masked load of exactly `len` bytes (no over-read). Digit placement is
-// computed from compressed delimiter positions; octet > 255 is a dword
-// compare on the zero-padded reversed digit group, in parallel with the
-// convert. Unusual-but-valid forms (octal, hex, leading zeros, fewer than
-// four parts) return ipv4_fast_fail so the general parser runs.
+// Needs VBMI2 for vpcompressb; ADA_AVX512 stays BW+VL (IPv6 does not use
+// VBMI2). Masked load of exactly `len` bytes (no over-read). Digit
+// placement is computed from compressed delimiter positions; octet > 255
+// is a dword compare on the zero-padded reversed digit group, in parallel
+// with the convert. Unusual-but-valid forms (octal, hex, leading zeros,
+// fewer than four parts) return ipv4_fast_fail so the general parser runs.
 ada_really_inline uint64_t try_parse_ipv4_avx512(const char* data,
                                                  size_t len) noexcept {
   // One trailing dot is WHATWG-legal ("1.2.3.4."); the SIMD kernel is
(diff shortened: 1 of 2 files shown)


lemire:
@anonrig Done with my friend grok.

include/ada/checkers-inl.h
```diff
@@ -125,40 +124,95 @@ parse_ipv4_decimal_scalar(const char* p, const char* pend) noexcept {
...
+  const __m128i zero = _mm_set1_epi8('0');
+  const __m128i digits = _mm_sub_epi8(v, zero);
+  const __mmask16 is_digit = _mm_cmple_epu8_mask(digits, _mm_set1_epi8(9));
+  // Junk in a digit slot: inside [0, len) yet neither a digit nor a dot.
+  const __mmask16 hole = _kandn_mask16(_kor_mask16(delim, is_digit), len_k);
+  const __m128i v0 = _mm_maskz_mov_epi8(is_digit, digits);
+
+  const __m128i iota =
+      _mm_setr_epi8(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+  const __m128i c = _mm_maskz_compress_epi8(delim, iota);
```
anonrig, line 168: `vpcompressb` needs `__AVX512VBMI2__`. `-march=skylake-avx512` / `-mavx512bw -mavx512vl` no longer compile. Don’t widen `ADA_AVX512` (IPv6 doesn’t need VBMI2).

include/ada/checkers-inl.h
```diff
@@ -125,40 +124,95 @@ parse_ipv4_decimal_scalar(const char* p, const char* pend) noexcept {
...
+      _kor_mask16(_kor_mask16(hole, static_cast<__mmask16>(over)), bad_gap);
+  const uint32_t gerr = static_cast<uint32_t>(_mm_popcnt_u32(dots) ^ 3u) |
+                        (zero_bits & start_bits & (keep >> 1));
+
+  if ((gerr | static_cast<uint32_t>(kerr)) == 0) [[likely]] {
+    uint8_t oct[4]{};
+    _mm_mask_cvtepi32_storeu_epi8(oct, 0x0F, res);
+    return (static_cast<uint64_t>(oct[0]) << 24) |
+           (static_cast<uint64_t>(oct[1]) << 16) |
+           (static_cast<uint64_t>(oct[2]) << 8) | oct[3];
```
anonrig: `oct[4]` forces a stack frame. `_mm_cvtepi32_epi8(res)` + `bswap` is a leaf and ~15% faster here.

anonrig (commented):
Looks correct. Two nits: gate this kernel on `__AVX512VBMI2__` (`vpcompressb` is not BW+VL), and pack with `vpmovdb`+`bswap` instead of `oct[4]` (drops the stack frame; ~15% faster at the kernel).

anonrig (approved):

include/ada/checkers-inl.h
lemire: I don't recommend using AVX-512 on anything before Intel Ice Lake. Intel messed it up and had something they call 'licensing' (the frequency goes down).

lemire: Addressed in 111dda98: the IPv4 kernel is now `ADA_AVX512 && __AVX512VBMI2__`. `ADA_AVX512` stays BW+VL so IPv6 is unchanged. `-march=skylake-avx512` / `-mavx512bw -mavx512vl` compile again and take the scalar IPv4 path. VBMI2 is Ice Lake / Zen 4 and later, so this also keeps AVX-512 IPv4 off pre-Ice Lake parts.

include/ada/checkers-inl.h
lemire: Addressed in 111dda98: octets are packed with `_mm_cvtepi32_epi8(res)` + `bswap` (`vpmovdb` + `bswapl`) instead of `oct[4]`.

include/ada/checkers-inl.h
lemire: Fixed but I recommend revising this policy, using AVX-512 on old Intel systems is unlikely to be a win.

PR #1234, opened 2026-08-26
RFC: url_pattern_list — compiled route-set matching (URLPatternList)

FranciscoThiesen:
## What this is

An RFC implementation of **set-level URLPattern matching** — `ada::url_pattern_list` — opened as a draft to gather feedback on whether this belongs in ada and in what shape, before I polish it further.

Routing over N routes with URLPattern today means a sequential `exec()` loop: O(N) per request, each step running regexes. Matteo Collina's ["You should not use URLPattern to route HTTP requests on the server"](https://adventures.nodeland.dev/archive/you-should-not-use-urlpattern-to-route-http/) documents the consequence, and the WHATWG fix — [`URLPatternList`, whatwg/urlpattern#166](https://github.com/whatwg/urlpattern/pull/166) — has been open since 2022 with no implementation anywhere. Since ada's URLPattern is what Node 24 and Workers ship, ada seems like the right place for the first one.

## What it does

`url_pattern_list::create(patterns)` compiles the **whole route set** once:

- Patterns are parsed by **ada's existing URLPattern machinery** (so canonicalization and part semantics are inherited, not re-invented); parts classify into literal / `:param` / `*` segments.
- The safe subset (which is ~all real route tables) compiles into a segment trie with per-node dispatch tables computed at build time — the same trick ada already hand-rolls for schemes (`(2 * scheme.size() + scheme[0]) & 7`), generalized: the builder *searches* for the few bytes that distinguish each node's children, plus two levels of whole-route lookup tables for fully-static and fixed-shape parameterized routes.
- Routes with regex groups, and inputs beyond the fast-path limits, fall back to a sequential path with identical semantics — correctness never depends on the fast path.
- `match(pathname)` is allocation-free: segment scan, table probes, bounded trie walk, captures returned as slices.

Two properties worth calling out:

1. **No regex engine needed for the safe subset** — a compiled list of static/`:param`/`*` routes never touches a regex provider. Given that ada currently has to label its only bundled provider *unsafe*, a routing path with no provider requirement at all seemed worth having.
2. **Bounded worst-case matching** — the walk is backtracking-bounded by construction (≤ 2·segments + constant node visits), i.e. the complexity-attack story for routing becomes a structural property instead of a regex-engine property.

## Numbers

From the self-contained benchmark included in this PR (`benchmarks/`), ~100 realistic REST routes, Apple M3 Max, LLVM 22:

```
routes: 101 / urls in stream: 512 (incl. ~20% near-misses, one live regexp route) / sequential-vs-list disagreements: 0
BasicBench_SequentialURLPatternExec   58.42 us/url   (17.1k urls/s)
BasicBench_URLPatternListMatch        54.00 ns/url   (18.5M urls/s)     ~1,082x
```

In a larger external harness (100–500 routes, 73k verified URLs, cross-checked against find-my-way and a naive matcher for identical answers), the same engine measures **5–35 ns/match** — ~7–23× find-my-way and 2,000–16,700× a sequential `exec()` loop. Happy to share that harness or run any benchmark you'd prefer.

## Open questions (the feedback I'm after)

1. **Is this desirable in ada at all** — or would you rather see it as a separate library on top of ada?
2. **Match-order semantics.** This implements specificity order (static > `:param` > `*` per segment, insertion-order tiebreak — find-my-way/Express-compatible). The spec PR stalled partly on exactly this question; insertion-order-first-match is implementable but forfeits some of the table shortcuts. Which contract should ada expose?
3. **API shape/naming** — `url_pattern_list` mirrors the WHATWG proposal; open to whatever fits ada's surface. Related: captures are currently returned as (offset,length) slices of the input (allocation-free), and regexp-route group values are not materialized (the caller re-execs that route's `url_pattern`) — is that the right trade for ada?
4. **Scope** — v1 is pathname-only (other components wildcard). Reasonable for an RFC?
5. The builder is deliberately constexpr-friendly — with C++26 the whole table set can freeze into `.rodata` (zero init; I keep that tier in a companion since ada is C++20). If there's interest I can sketch how it composes with the `more-constexpr` direction.

Limits, design notes, and tests (including a randomized differential test against an independent reference matcher) are in the code. If the direction is wrong, telling me so bluntly saves us both time — that's what the draft status is for.


lemire:
There are compile-time regex engines so supporting some flavor of regex is 'easy'. 

The application is obviously fantastic. You could possibly be *massively* faster with a compile-time expression.

But my impression is that URL Patterns are meant and used in a dynamic context.

anonrig:
I'm not against it but this needs to be:
- tested (500 line tests is not enough)
- fuzzed (there is no fuzzer change in this PR)
- the API surface area should be guarded with private doc comments - other than public API.
- needs to be communicated with the WHATWG changes https://github.com/whatwg/urlpattern/issues/30

FranciscoThiesen:
@lemire — agreed on the dynamic context, and that's what this PR ships: the "compilation" here happens at `create()` time in microseconds-to-milliseconds, so fully dynamic registration (Node-style) is the primary path. The C++-compile-time tier exists as a companion experiment — same builder run under `constexpr`, tables verified byte-identical, zero init — for deployments where the route set is fixed per build (Workers-style, embedded). Happy to sketch how it composes with the `more-constexpr` direction if there's interest. And yes on compile-time regex engines — a CTRE-style residual in the frozen tier is a lovely direction; in the dynamic tier I went the other way for the common case: the static/`:param`/`*` subset provably needs no regex engine at all, which seemed worth having given the provider situation.

@anonrig — all four addressed in ad28a3d:

- **Tests**: expanded coverage-first — built with llvm-cov and closed the uncovered branches Codecov flagged. Patch coverage on the new files now measures 99.7% line / 98.6% branch locally; the residual is a handful of defensive demotion paths behind offline-search failure that are not adversarially constructible (happy to detail). Added a ~130-case data-driven table, property tests cross-checking winners against per-route `ada::url_pattern` itself, and group-name/duplicate-param cases.
- **Fuzzer**: `fuzz/url_pattern_list.cc` following `fuzz/url_pattern.cc` conventions (FuzzedDataProvider, `.options`, wired into `build.sh`), with three strategies: arbitrary-pattern-set construction, a differential oracle (fast path vs reference matcher, `abort()` on disagreement so libFuzzer catches it), and fast-path-gate boundary crossing. Ran locally 10 min under ASan+UBSan: 372k execs, zero findings (one investigated non-finding: ASan allocator high-water from std::regex-heavy units, matching the existing url_pattern fuzzer's exposure; options mirror url_pattern.options).
- **Private API guarding**: internal machinery now carries `@private` doc comments per ada convention (and what could move out of the public header did).
- **WHATWG**: drafting a note for whatwg/urlpattern#30 referencing this implementation, the no-regex-engine data point, and the match-order question that stalled #166 — will post it there shortly.

On CodSpeed: the three flagged regressions are URL-setter benchmarks (`SetHash`/`SetProtocol`/`SetPassword`) that this additive, `ADA_INCLUDE_URL_PATTERN`-gated diff doesn't touch — CodSpeed's own report flags "different runtime environments" for those comparisons. The new sequential-exec benchmark was also slimmed so simulator iterations stay cheap.

anonrig:
Can we stop the AI responses and just talk for a second...

lemire:
@FranciscoThiesen 

Ok, I misread the PR, so 'build time' is not 'compile time' here. 

Sorry, I did not read the code yet.

So I think that this might be more applicable then. This could be called from JavaScript... so you'd have your set of routes, the C++ code would not some magic and then dispatch them.


@anonrig This sounds like it could be a big deal if Node.js would accept it, no?

FranciscoThiesen:
> Can we stop the AI responses and just talk for a second...

"Please reply to Yagiz and try not to sound like AI from now on. Make no mistakes!", just kidding (:

Sure! Let me know what else you want to know about the PR. Will post on the WHATWG thread later today, want to make sure I have full context on the thread.

@anonrig @lemire  Feel free to push back on any aspects of the PR, as I am a tourist here...

FranciscoThiesen:
https://github.com/whatwg/urlpattern/issues/30 @anonrig @lemire here is my reply on the correlated WHATWG issue

anonrig:
@jasnell are you interested in adding this to cloudflare workers?

jasnell:
Not opposed but will have to evaluate. Won't be an immediate priority

include/ada/url_pattern_list.h
```diff
@@ -0,0 +1,505 @@
...
+   * route id reported by
+   * match(). Duplicate patterns are allowed; the smaller index wins.
+   * All inputs must be valid UTF-8.
+   *
+   * @return The compiled list, or an error when a pattern is not a valid
+   *         URLPattern pathname pattern (errors::type_error, matching the
+   *         URLPattern constructor's behavior).
+   */
+  static tl::expected<url_pattern_list, errors> create(
+      const std::vector<std::string_view>& pathname_patterns);
```
anonrig: The `regex_provider` template is the right idea — same as `url_pattern`.

This still isn’t the same option we give callers of URLPattern:

- `parse_url_pattern<Provider>(input, base, options)` is the public entry point. There is no `parse_url_pattern_list<Provider>(...)` next to it.
- `create()` does not take `url_pattern_options*` (`ignore_case`).
- `create()` only accepts pathname strings. Node/Workers already have `url_pattern<Provider>` objects with a compiled engine; there is no way to pass those in.

Please match the `parse_url_pattern` shape: provider as a template, plus `url_pattern_options`.

include/ada/url_pattern_list.h
```diff
@@ -0,0 +1,505 @@
...
+  std::array<uint8_t, 4> witness_ids{};
+  // Absolute segment position of each witness.
+  std::array<uint8_t, 4> witness_positions{};
+  // Alphabet id of each witness.
+  std::array<uint8_t, 4> witness_sub_ids{};
+  std::array<uint8_t, max_captures_per_route + 1> run_first{};
+  std::array<uint8_t, max_captures_per_route + 1> run_last{};
+  // Segment positions of the params, ascending.
+  std::array<uint8_t, max_captures_per_route> param_positions{};
+};
```
anonrig: `mask`, `n_static`, and `witness_ids` are written at build time and never read by the matcher. `witness_positions` / `witness_sub_ids` are what `shape_lookup` actually uses. Don’t keep fields the hot path never touches.

include/ada/url_pattern_list.h
```diff
@@ -0,0 +1,505 @@
...
+
+/**
+ * Internal machinery for ada::url_pattern_list: the provider-independent
+ * route-set compiler and matcher. Nothing in this namespace is part of the
+ * supported public API except the limit constants and the `capture` struct
+ * (re-exported by url_pattern_list_match_result), which are documented
+ * because they delimit the fast path and the result contract. Everything
+ * else lives here only because the url_pattern_list class template needs it
+ * and is marked @private; it may change at any time.
+ * @namespace ada::url_pattern_list_helpers
```
anonrig: This is all in the public header (`ada.h` includes it). `@private` comments are not enough — anyone including ada sees `trie_node`, `shape_group`, `compile_route_set`, etc.

Public surface should be `url_pattern_list`, the match result, and the limit constants. The compiler types can live in a header the fuzzer includes, not in `ada.h`.

include/ada/url_pattern_list-inl.h
```diff
@@ -0,0 +1,158 @@
...
+  }
+  if (candidate.mode == helpers::route_mode::regexp) {
+    const url_pattern_component<regex_provider>& component =
+        regexp_components_[static_cast<size_t>(candidate.regexp_component)];
+    if (component.fast_test(pathname)) {
+      // Capture slices are not recoverable through the provider interface;
+      // see url_pattern_list_match_result.
+      best = helpers::engine_result{};
+      best.route = static_cast<int32_t>(route_index);
+    }
```
anonrig: The provider already has `regex_search`. Don’t drop captures here and tell the caller to re-exec a `url_pattern`.

`url_pattern` uses `regex_provider::regex_search` for `exec`. Do the same. `fast_test` / `regex_match` only answers yes/no.

include/ada/url_pattern_list-inl.h
```diff
@@ -0,0 +1,158 @@
...
+                                route.group_names)) {
+      helpers::finalize_route(route);
+    } else {
+      // The pattern needs URLPattern regexp semantics: compile it as a
+      // pathname component through the provider; it participates in the
+      // priority order via its approximated kind sequence.
+      auto compile_options = url_pattern_compile_component_options::PATHNAME;
+      auto component = url_pattern_component<regex_provider>::compile(
+          list.patterns_[i], url_pattern_helpers::canonicalize_pathname,
+          compile_options);
```
anonrig: `url_pattern_component::compile` takes `url_pattern_compile_component_options`, which has `ignore_case`. This always compiles with the default (case-sensitive).

`parse_url_pattern` forwards `url_pattern_options::ignore_case` into the component. This path should too, otherwise a custom provider never sees the flag.

src/url_pattern_list.cpp
```diff
@@ -0,0 +1,1802 @@
...
+    uint64_t x, y;
+    std::memcpy(&x, a + i, 8);
+    std::memcpy(&y, b + i, 8);
+    acc |= x ^ y;
+  }
+  uint64_t x, y;
+  std::memcpy(&x, a + len - 8, 8);
+  std::memcpy(&y, b + len - 8, 8);
+  return (acc | (x ^ y)) == 0;
+}
```
anonrig: Do we need a hand-rolled `eq_bytes`? `memcmp` is enough. Same question for the NEON segment scan below — the scalar loop is the fallback and is what every non-ARM host runs. Don’t add a second scanner unless it is required for the API.

src/url_pattern_list.cpp
```diff
@@ -0,0 +1,1802 @@
...
+
+// Witness-subset search for one shape group, over the per-static-segment
+// alphabet. Preferred features first (first byte, last byte, length of each
+// static), deeper bytes after; constant and duplicate feature columns are
+// pruned; subsets covering more distinct static segments are preferred
+// (they reject more non-matching inputs at the slot).
+bool find_shape_plan(const std::vector<transient_shape_entry>& entries,
+                     uint32_t n_static, std::array<uint8_t, 4>& ids_out,
+                     uint8_t& n_ids_out) {
```
anonrig: `n_ids_out` is written and never read by `compile_shape_group`. Dead out-parameter.

tests/url_pattern_list_tests.cpp
```diff
@@ -0,0 +1,1168 @@
...
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "gtest/gtest.h"
+
+#include "ada.h"
+
+using regex_provider = ada::url_pattern_regex::std_regex_provider;
+using list_type = ada::url_pattern_list<regex_provider>;
```
anonrig, line 14: Every test hardcodes `std_regex_provider`. `url_pattern` is specifically designed so Node/Workers can pass their own engine. Add one test with a different provider (even a thin wrapper around `std_regex_provider` that counts `create_instance` / `regex_search`) so we know the template is actually used.

anonrig (commented):
Two things before this grows further:

1. **Regex engine.** The `regex_provider` template is correct. It still does not match `parse_url_pattern`: no free function, no `url_pattern_options` / `ignore_case`, regexp routes use `fast_test` and drop captures instead of `regex_search`, and tests only ever instantiate `std_regex_provider`.
2. **Don’t add code we don’t use.** `shape_group::{mask,n_static,witness_ids}` are never read at match time, `n_ids_out` is a dead out-parameter, and the public header ships the whole compiler (`trie_node`, `shape_group`, `compile_route_set`, …). `eq_bytes` and the NEON scanner are extra unless we can show we need them.

Public API should stay `url_pattern_list` + match result + limits, with the provider passed the same way as URLPattern.

include/ada/url_pattern_list-inl.h
```diff
@@ -0,0 +1,158 @@
...
+    // leading '/'): the sequential fallback matches every route with
+    // identical priority semantics.
+    return match_sequential(pathname);
+  }
+  // Routes the compiled tables cannot answer for still participate in the
+  // priority order: the winner is decided by (kind sequence, insertion
+  // index), never by which path matched it.
+  for (const uint32_t route_index : auxiliary_routes_) {
+    consider_route(route_index, pathname, best);
+  }
```
anonrig: This loop always runs, even after an exact-table hit. A full-static winner already has the best kind sequence — almost nothing can outrank it.

At `create()`, record which auxiliary routes can actually beat each compiled winner (usually none, or a couple of regexp routes). After an exact/shape hit, only test those.

This is the first thing to fix if the table has even one `(\\d+)` route. `std::regex` on the leftover list will dominate the 50ns walk.

src/CMakeLists.txt
```diff
@@ -64,6 +64,12 @@ endif()
 # inlining budget for setters matches main.
 target_compile_definitions(ada PRIVATE ADA_PERCENT_ENCODE_SIMD_SEPARATE_TU=1)
 
+# Keep the URLPattern route-set compiler in its own TU for the same reason:
+# ~1800 extra lines in the unity TU reshuffle GCC's unit-wide inlining budget
+# and de-inline url_aggregator setter hot paths (CodSpeed: SetHash -12%).
+# The amalgamated single-file build still includes it inline.
+target_compile_definitions(ada PRIVATE ADA_URL_PATTERN_LIST_SEPARATE_TU=1)
```
anonrig: Right split for GCC setter inlining, wrong split for a 50ns match. `match_compiled` / `scan_segments` / `dispatch_static` cannot inline into the caller from this TU.

Keep `compile_route_set` here. Put the walk in a header as `ada_really_inline`.

src/url_pattern_list.cpp
```diff
@@ -0,0 +1,1802 @@
...
+    const transient_node& tn = t[static_cast<size_t>(order[ni])];
+    const size_t nk = tn.kids.size();
+    if (nk == 0) {
+      nd.dispatch = 0;
+      continue;
+    }
+    if (nk <= 2) {
+      nd.dispatch = 1;
+      continue;
+    }
```
anonrig, line 417: Projection at 3 children, then you still `verify_edge`. For `/users`, `/posts`, `/health` that is gather + multiply + slot + compare, vs two or three 4–8 byte compares.

Raise the direct threshold to ~8 and measure. Keep the hash path for a wide root node.

include/ada/url_pattern_list.h
```diff
@@ -0,0 +1,505 @@
...
+  uint16_t static_witness_plan = 0;
+  uint8_t static_slot_bits = 0;
+  uint8_t has_static_table = 0;
+  // Shape tables: whole-route witness dispatch for parameterized routes.
+  std::vector<shape_group> groups{};
+  std::vector<shape_directory_entry> group_directory{};
+  std::vector<uint8_t> group_slots{};
+  std::vector<static_entry> group_entries{};
+  uint8_t has_shape_tables = 0;
+};
```
anonrig, line 257: Ten `std::vector`s. A miss of the exact table touches nodes, edges, slots, blob, `route_meta`, then groups.

Pack the tables into one arena so the walk stays in a couple of cache lines.

include/ada/url_pattern_list.h
```diff
@@ -0,0 +1,505 @@
...
+  uint16_t n_static = 0;
+  // Witness plan: 3 nibble ids | count << 12 | use_len << 14.
+  uint16_t witness_plan = 0;
+  // 0 none, 1 direct (<=2 children), 2 projection, 3 linear scan.
+  uint8_t dispatch = 0;
+  // Slot table size == 1 << slot_bits.
+  uint8_t slot_bits = 0;
+  // 1 if the node still has a param/wildcard alternative to backtrack to.
+  uint8_t has_alternative = 0;
+};
```
anonrig: ~40 bytes, and most nodes are dispatch 0/1 — they never use `multiplier` / `witness_plan` / `slot_base` / `slot_bits`. Store that payload only on hashed nodes.

src/url_pattern_list.cpp
```diff
@@ -0,0 +1,1802 @@
...
+  }
+#endif
+  if (overflow || nseg >= max_fast_path_segments) {
+    return 0;
+  }
+  soff[nseg++] = static_cast<uint16_t>(s);
+  soff[nseg] = static_cast<uint16_t>(ulen + 1);  // sentinel
+  return nseg;
+}
```
anonrig: Typical pathnames are 20–40 bytes with 2–4 slashes. NEON setup + a `ctz` loop is for long inputs.

A portable SWAR slash scan will beat this on the common case. Keep NEON only if a 1k+ pathname benchmark says so.

src/url_pattern_list.cpp
```diff
@@ -0,0 +1,1802 @@
...
+    if (ulen >= 8) {  // one unaligned load, not a byte loop
+      const uint64_t t8 = load8_le(url + (ulen - 8));
+      tail_off = ulen - 8;
+      std::memcpy(endpad, &t8, 8);
+    } else {
+      for (uint32_t j = 0; j < ulen; j++) {
+        endpad[j] = url[j];
+      }
+    }
+  }
```
anonrig: `verify_edge` does 8-byte loads, so almost every trie walk builds `endpad` — last segments are usually short (`42`, `me`).

Byte-compare tails < 8 instead of padding. Then this goes away.

src/url_pattern_list.cpp
anonrig: Fine for 4–8 bytes. Past 16, libc `memcmp` is already SIMD — and the exact-table compare is the one that sees full pathnames.

src/url_pattern_list.cpp
```diff
@@ -0,0 +1,1802 @@
...
+    default: {  // linear demotion rung (also carries fanout > 254)
+      for (uint32_t j = 0; j < nd.n_static; j++) {
+        if (verify_edge(r.edges[static_cast<size_t>(nd.first_child) + j], s,
+                        r.blob.data())) {
+          return static_cast<int32_t>(j);
+        }
+      }
+      return -1;
+    }
+  }
```
anonrig: Root usually has the highest fanout. A 256-entry first-byte table (or a map of the first segment) jumps to the child without a witness plan. find-my-way does this; it is simpler than hashing the whole pathname.

anonrig (commented):
Performance notes on the match path. The walk is already in the tens of nanoseconds; the remaining time is extra work after a hit, and layout.

Highest impact:
1. Don’t walk `auxiliary_routes_` after a decisive exact/shape hit — especially if any route is regexp (`std::regex` will dominate).
2. Inline the matcher; keep only the builder in the extra TU.
3. Direct-compare small fanouts (~8) instead of projecting at 3 children.
4. Pack the ~10 table vectors into one arena; shrink `trie_node` so dispatch 0/1 nodes don’t carry hash payload.
5. SWAR slash scan + byte-compare short tails (drop `endpad` / NEON unless a long-pathname bench says otherwise). `memcmp` past 16 bytes. First-byte index at the root.

I would not add more shape-table machinery or constexpr tables for the Node path. The provider template is the speed knob for regexp routes — V8/JIT, not `std::regex`.

include/ada/url_pattern_list.h
FranciscoThiesen: Fixed.

- `parse_url_pattern_list<Provider>(std::span<const std::string_view>, const std::string_view* base_url = nullptr, const url_pattern_options* options = nullptr)` is declared in `implementation.h` next to `parse_url_pattern` and defined in `implementation-inl.h` like it. With a base URL, each pattern is processed the way the pathname of a URLPatternInit is.
- `parse_url_pattern_list<Provider>(std::span<const url_pattern<Provider>>)` takes existing objects. It reads the pattern from `get_pathname()` and the flag from `ignore_case()`. Routes that need the engine reuse the object's compiled `pathname_component`, so there is no second `create_instance`.
- `create()` is private now.

One thing I ran into: `parse_url_pattern_impl` forwarded `options->ignore_case` into the component options but never stored it on the `url_pattern`, so `url_pattern::ignore_case()` always returned false. (also fixed in my PR)

include/ada/url_pattern_list.h
FranciscoThiesen: Removed, along with the shape tables themselves, about 500 lines: with the first-byte index at the root, static hits measure about the same (26.8 -> 28.2 ns/url) and `:param` hits cost about 4 ns more (30.5 -> 34.6). Details on the `src/url_pattern_list.cpp:380` comment.

include/ada/url_pattern_list.h
FranciscoThiesen: Moved.

- `include/ada/url_pattern_list.h` keeps the limits, the match result, `url_pattern_list`, and the table records the inline matcher reads, under `url_pattern_list_detail`. The walk is inline per your other comment, so those records have to be visible.
- The compiler (`route_info`, `classify_parts`, `compile_route_set`, the witness planners, the arena packing) is in `src/url_pattern_list_compiler.h`. `ada.h` does not include it and it is not installed.
- The fuzzer gets it through the amalgamated `ada.cpp`, like the other fuzzers. The OSS-Fuzz script only has `-I build/singleheader`.

One more thing in that header since: `url_pattern_list-inl.h` includes `<arm_neon.h>` under `ADA_NEON`, now that the NEON scan is back. The `src/` files did already; this is the first public header that does. Say if that bothers you.

include/ada/url_pattern_list-inl.h
FranciscoThiesen: Changed. A regexp route now runs `regex_match` and then `regex_search`, the same test/exec split `url_pattern` has, and the result carries the groups: `regexp_route` plus `regexp_groups`, aligned with `group_names()`. Subset routes keep the `(offset, length)` slices.

I kept the `regex_match` step in front on purpose. With libc++ a `regex_search` miss on a `(\d+)` route scans from every start position, so a near miss costs far more than a failed `regex_match`. With the pruning from your other comment the provider is rarely reached at all, so this only matters for the routes that do get tested, but there it is the difference between a cheap rejection and a scan.

include/ada/url_pattern_list-inl.h
FranciscoThiesen: Fixed: `compile_options.ignore_case = options->ignore_case` for the regexp components. The counting-provider test checks that the flag reaches `create_instance`.

The static / `:param` / `*` subset folds ASCII to match. Literals are folded at creation, and the input is folded into a stack copy at match time, so captures still slice the original. There is a parity test against `url_pattern` with the option set.

src/url_pattern_list.cpp
FranciscoThiesen: No we don't, I've removed it.

A segment compare is now:

- 8 to 16 bytes: two 8-byte loads inside the segment
- under 8 bytes: one masked load when 8 bytes are readable, a byte gather otherwise
- over 16 bytes: `memcmp` against the blob

The NEON scan I kept, though. You asked for a 1k+ pathname benchmark before keeping it (your comment on `src/url_pattern_list.cpp:338`), and it came out faster at every length, not just on long inputs; the numbers are there. The SWAR loop is the portable path and also handles inputs under 16 bytes on ARM.

src/url_pattern_list.cpp
FranciscoThiesen: Removed.

tests/url_pattern_list_tests.cpp
FranciscoThiesen: Added `counting_provider`, a wrapper around `std_regex_provider` that counts `create_instance`, `regex_search` and `regex_match` and records the `ignore_case` flag it was given. The tests that instantiate `url_pattern_list<counting_provider>` assert exact counts:

- one `create_instance` per regexp route, none for subset routes
- one `regex_match` plus one `regex_search` on a regexp hit
- zero provider calls after a subset hit that nothing can outrank

include/ada/url_pattern_list-inl.h
FranciscoThiesen: Thanks for pointing this out, it was the biggest win.

At creation each trie route now records which auxiliary routes both outrank it and could match the same input (literal prefixes agree, segment counts are compatible, `:param` never binds an empty segment). After a hit only those run, and for most routes the list is empty. On top of that, a regexp route made of fixed text and `:name` groups has an exact segment shape, and one with custom groups an anchored literal prefix; that is checked before any provider call, on hits and on misses.

The table with one `(\d+)` route went from 110.7 to 42.5 ns/url. On that stream the regex now runs once in 32 URLs (an `/api/v1/invoices/.../zz` miss that lands on `/*`, where the regexp route legitimately could win) instead of on every URL. Test: `auxiliary_routes_are_pruned_after_a_fast_path_hit`.

src/CMakeLists.txt
FranciscoThiesen: Moved. `scan_segments`, `verify_edge`, `dispatch_static` and `match_compiled` are `ada_really_inline` in `url_pattern_list-inl.h`. Only the builder stays in `src/url_pattern_list.cpp`.

I verified the GCC 14 `-O3` output of the unity `ada.cpp` against main: all 793 functions, `url_aggregator` setters included, still have identical opcode streams.

src/url_pattern_list.cpp
FranciscoThiesen: I've set to 8, after measuring where the projection starts to win. One node with N static children plus a `:rest` sibling, inputs spread evenly over the children. Keys are either normal names of mixed length (`users`, `orders`, ...) or all the same length and prefix (`seg` + three letters, so the length check rejects nothing). ns/url, median of 3 runs, projection / direct:

| N | hits, mixed | hits, same length | misses, mixed | misses, same length |
|---:|---|---|---|---|
| 3 | 20.5 / 19.0 | 16.8 / 21.4 | 20.6 / 20.3 | 17.5 / 18.9 |
| 8 | 20.6 / 19.7 | 16.8 / 21.8 | 20.7 / 23.7 | 17.5 / 25.5 |
| 12 | 21.1 / 20.4 | 16.8 / 24.9 | 21.2 / 26.1 | 17.5 / 31.0 |
| 16 | 21.0 / 21.7 | 16.9 / 27.1 | 21.1 / 28.9 | 17.5 / 36.5 |
| 24 | 20.7 / 23.6 | 16.8 / 31.1 | 20.7 / 34.1 | 17.5 / 47.4 |

With mixed lengths your intuition is right: 2-3 compares win by ~1.5 ns at 3 children, ~1 ns at 8, even by 12-16. With same-length keys the projection wins at every N (4-5 ns on hits, more on misses), presumably because the loop's exit branch mispredicts once the length check stops filtering. 8 keeps the common case and bounds the other; on the PR table it's a wash against projection from 3 (28.0 vs 28.3 static, 34.9 vs 35.0 `:param`).

I had 16 at first: +2.5 ns on the PR table, but only because that stream is skewed to `/api` and the exit becomes predictable. The sweep I ran is about 120 lines on google benchmark; I can push it or gist it if you want.

include/ada/url_pattern_list.h
FranciscoThiesen: Done. It is one arena now, 8-byte-aligned sections addressed by offset: nodes, hash payloads, edges, slots, blob, route records, segment table, aux table, root index. The walk derives its typed view once per match.

include/ada/url_pattern_list.h
FranciscoThiesen: Ok!

Now `node_record` is 24 bytes. `multiplier`, `slot_base`, `witness_plan` and `slot_bits` moved to a `hash_record` that only projection nodes reference.

src/url_pattern_list.cpp
FranciscoThiesen: Gone. A tail under 8 bytes compares through one masked load when 8 bytes are readable from the segment start, and through a byte gather otherwise. `short_tail_segments_compare_exactly` covers tails of 1 to 7 bytes, last-byte mismatches, and one byte too many or too few.

src/url_pattern_list.cpp
FranciscoThiesen: I ran the 1k+ benchmark before deciding. Two-node table where `/files/*` wins, so the scan dominates; NEON is the scanner from before the refactor, SWAR the portable loop that had replaced it. ns/url:

| pathname | NEON | SWAR |
|---|---:|---:|
| 32 B, 2 segments | 15.4 | 18.7 |
| 128 B, 2 segments | 17.3 | 26.8 |
| 1 KB, 2 segments | 40 | 132 |
| 4 KB, 2 segments | 126 | 460 |
| 128 B, 24 segments | 24.1 | 37.2 |
| 1 KB, 24 segments | 47 | 137 |
| 4 KB, 24 segments | 131 | 585 |

So it does say so, and not only past 1 KB: NEON wins at every length on the M3, 1 to 3 ns on 20 to 40 byte paths.

So it stays. The refactor had dropped it; the latest push puts it back under `ADA_NEON` for inputs of 16 bytes or more (16 bytes per step, the `/` compare narrowed to a nibble mask, an overlapped last block so nothing is read past the input), with the SWAR loop as the portable path and for shorter inputs. The sweep test runs both against a naive scanner over lengths 1 to 70 and every slash placement. 32 bytes is back at parity (15.6 vs 15.8). What is left on long `*` tails, 2x at 4 KB, is the line-terminator check for `(.*)`, a second pass over the tail; it could fold into the scan if that ever matters.

Building this also caught a bug of mine: the tail check for `*` routes walked the tail byte by byte, 1.3 us at 4 KB. It is a SWAR pass now.

src/url_pattern_list.cpp
FranciscoThiesen: Added. Root children are sorted by first byte, and a 256-entry table jumps to the run of children that share it. Runs of up to 8 compare directly; longer runs fall back to the ladder.

With that in place I measured the whole-pathname exact table and the shape tables against the plain walk, ns/url, before -> after:

| stream | with the tables | without |
|---|---:|---:|
| static hits | 26.8 | 28.2 |
| `:param` hits | 30.5 | 34.6 |

Static hits are within run-to-run noise, so the exact table is gone. The shape tables were worth about 4 ns on `:param` hits; I removed them too, about 500 lines, since a single trie walk is what you asked to keep simple. If you would rather have the 4 ns, the shape table is one probe in front of the walk and can come back on its own.

src/url_pattern_list.cpp
FranciscoThiesen: Past 16 bytes the compare is `memcmp` against the blob; 8 to 16 bytes is two whole-word loads, under 8 one masked load or a byte gather. The exact table, the one that compared whole pathnames, no longer exists: with the first-byte index at the root plus direct compares, static hits measure 26.8 -> 28.2 ns/url before -> after, about a nanosecond, so I dropped it, and the shape tables with it, about 500 lines.

anonrig:
One thing we should make sure before landing is:

- We should use url_pattern_regex_provider as a templated argument for this rather than implementing a regex parser etc. This is required because Node.js and other solutions might want to use v8 rather than any other solution.

anonrig:
Can you also add a documentation to the readme, github.com/ada-url/website as well?

FranciscoThiesen:
> One thing we should make sure before landing is:
> 
> * We should use url_pattern_regex_provider as a templated argument for this rather than implementing a regex parser etc. This is required because Node.js and other solutions might want to use v8 rather than any other solution.

It already is. `url_pattern_list` is templated on the provider exactly like `url_pattern`, under the same `regex_concept`, and the only regex calls it makes are the provider's `regex_match` and `regex_search` through `url_pattern_component`. 

There is no regex parsing in the list: custom `(...)` groups are opaque parts, and the whole component goes to the provider. Node's `URLPatternRegexProvider` satisfies the concept as is, so `ada::url_pattern_list<URLPatternRegexProvider>` compiles against V8 unchanged. The static / `:param` / `*` routes never touch a regex engine, which the counting-provider tests check (zero provider calls on those hits).

anonrig (approved):

anonrig:
CI is failing

FranciscoThiesen:
@anonrig the CI failures are fixed, they were both mine:

- macOS 14 (Apple clang 15) rejected the friend declaration that redeclares the constrained function template, so `parse_url_pattern_list` could not reach the private members. The overloads are now declared before the class and it befriends those specializations.
- MSVC failed to link because `ada_never_inline` there is `__declspec(noinline)` with no inline linkage, so the wildcard tail check was emitted in every translation unit. It moved to `url_pattern_list.cpp` next to the other detail functions; it was never inlined anyway.

Locally: 403 tests pass, the fuzzer is clean over 96k runs under ASan and UBSan, and the GCC unity build is still byte for byte identical to main on all 795 functions, setters included.

The runs on the last push are waiting for approval, so nothing has reported yet. Could you enable them when you get a chance? I have also resolved the review threads.

anonrig:
ci is still failing

FranciscoThiesen:
Looks like a flake to me. The failure is gtest's 5 second test discovery timing out, and it hit url_components and wpt_url_tests too, which this PR does not touch. The same job passed on the two previous pushes, clangcl Release passed on this one, and that Debug job ran about twice as slow as it did three days ago on the same image.

Can you re-run it?

FranciscoThiesen:
CI is now green @anonrig

PR #1235, opened 2026-08-27
enforce max_input_length on parse_url_impl early returns

Arawoof06:
parse_url_impl only re-checks the normalized url against max_input_length at the end of the state machine, but the query state and the authority and path-start early returns append a percent-encoded query or fragment and return before reaching it. a url whose input is under the limit but whose normalized href is not is therefore accepted when the parse lands on one of those slow-path returns (file://x/?<spaces>y, http://u@x#<spaces>y, http://[::1]#<spaces>y) while the equivalent https://x/?<spaces>y that takes the absolute fast path is rejected, so validity depends on which terminal state the parse happens to hit. this hoists the existing end-of-parse size check into an enforce_max_input_length helper and runs it on those early returns too. added typed regression tests over url and url_aggregator for the query and fragment paths.

### DIFF
--- a/src/parser.cpp
+++ b/src/parser.cpp
@@ -1851,6 +1851,29 @@ result_type parse_url_impl(std::string_view user_input,
 
   const uint32_t max_input_length = ada::get_max_input_length();
 
+  // Normalization (percent-encoding, IDNA) can push the serialized URL past
+  // max_input_length even when the input fit under it. The check at the end of
+  // the state machine covers URLs that run to completion, but several states
+  // return early after appending a query or fragment, so run the same check on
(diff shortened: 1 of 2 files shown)


src/parser.cpp
```diff
@@ -1851,6 +1851,29 @@ result_type parse_url_impl(std::string_view user_input,
 
   const uint32_t max_input_length = ada::get_max_input_length();
 
+  // Normalization (percent-encoding, IDNA) can push the serialized URL past
```
lemire, line 1854: Please trim these comments. It is unnecessary.

anonrig (approved):

PR #1242, opened 2026-08-29
test: cover the "::" grammar boundaries of the IPv6 host parser

ikkyu3:
## Summary

Adds regression tests for host-parsing boundaries around `::` in IPv6
literals — group count, per-group hex width, and lone (non-doubled)
leading/trailing colons — that aren't currently exercised by
`tests/wpt/urltestdata.json`'s existing IPv6 cases.

Motivated by #1233, which rewrote this parser; the new tests pin down the
exact WHATWG grammar boundaries it (and the scalar fallback) must continue
to enforce, independent of which code path handles a given build.

## What's covered

- `::` must elide at least one group (all-8-groups-explicit + `::` is invalid)
- a single elided zero-piece is *not* re-compressed to `::` on serialization
  (only runs of ≥ 2 are, per spec)
- a group holds at most 4 hex digits
- uppercase hex digits are accepted and normalized to lowercase
- a lone leading or trailing colon is invalid even when a genuine `::`
  appears elsewhere in the same host
- more than one `::` is invalid
- the full-width, no-compression boundary (8 groups × 4 hex digits, the
  longest valid non-compressed address)

## Test plan

- [x] New tests pass for both `ada::url` and `ada::url_aggregator`
- [x] Full `basic_tests` suite passes locally (207/207)
- [x] `clang-format` clean (LLVM 22, project config)

### DIFF
--- a/tests/basic_tests.cpp
+++ b/tests/basic_tests.cpp
@@ -1043,6 +1043,60 @@ TYPED_TEST(basic_tests, ipv4_fast_path_malformed_groups) {
   SUCCEED();
 }
 
+// Regression coverage for the "::"-bearing branch of the IPv6 host parser
+// (include/ada/url_ip-inl.h, try_parse_ipv6_avx512's `dc` branch on
+// AVX512VBMI2 builds; the scalar compress-shift loop in url.cpp otherwise).
+// These specifically target invariants enforced only by that branch's own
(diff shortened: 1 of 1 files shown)


lemire:
Running tests.

anonrig (approved):

PR #1251, opened 2026-09-14
apply the bidi rule to every label of a bidi domain name in to_ascii

Arawoof06:
is_label_valid only runs the RFC 5893 Bidi conditions on a label that has an RTL code point of its own, so the other labels of a Bidi domain name are never held to them and to_ascii accepts hosts such as ا.1x, x.1.א and ا.a- that UTS 46 (and whatwg-url, tr46, ICU) reject; only the URL standard's ASCII carve-out keeps xn--mgb.1x valid. to_ascii now classifies the mapped domain up front and validates every label with that flag, restarting the walk once when the RTL code points turn out to sit inside an ACE label of a non-ASCII domain. The same change is ada-url/idna#91, which this file is vendored from.


lemire:
I encourage you to submit the PR to idna and we then get the fix from idna.

It is too messy if we change everything at once.

lemire:
@anonrig I recommend closing this PR, not that the change is not welcome, but just because it is duplicated effort.

